PR #24, opened 2017-02-19
MIA PT Sherwin Garcia

sherwino:
Wasn't that complicated, if I opened the test.js to reveal what was being tested. 
Jolly good show mate.

### DIFF
--- a/starter-code/lib/viking.js
+++ b/starter-code/lib/viking.js
@@ -19,13 +19,96 @@
 //   FootViking: FootViking
 // };
 
-function Soldier () {}
 
-function Viking () {}
 
-function Saxon () {}
+function Soldier (name) {
+  this.name = name;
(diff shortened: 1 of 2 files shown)


starter-code/lib/viking.js
```diff
@@ -19,13 +19,96 @@
...
+  this.receiveDamage = function (damage) {
+    this.health = this.health - damage;
+    return "ouch";
+
+  };
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
+
```
nickborbe: If you use `.call` on Soldier here, then you get health and strength from soldier so there's no point in passing them in as arguments since they're getting set to 300 and 150 respectively.  Just FYI

nickborbe:
Nice work.

PR #26, opened 2017-02-19
MIA PT Orestes-Garcia

ogm710811:


starter-code/lib/viking.js
```diff
@@ -19,13 +19,87 @@
...
+    this.health = this.health - damage;
+    return this.health;
+  };
+}
 
-function Saxon () {}
+//------------------------------------------------------
+// VIKING
+//------------------------------------------------------
+function Viking (name, health, strength, damage) {
```
nickborbe: Don't need to pass in damage as an argument when you create a new Viking.  That's what the strength attribute is for.

starter-code/package.json
```diff
@@ -16,7 +16,8 @@
     "faker": "^3.0.1"
   },
   "devDependencies": {
-    "mocha": "^3.1.2",
+    "chai": "^3.5.0",
+    "mocha": "^3.2.0",
     "mocha-let": "^0.1.1"
   }
 }
```
nickborbe: Nice work.  Passing all the tests.

PR #27, opened 2017-02-20
MIA PT Adel Acebal

aacebal:


starter-code/lib/viking.js
```diff
@@ -19,13 +19,88 @@
 //   FootViking: FootViking
 // };
 
-function Soldier () {}
+function Soldier (name, health, strength) {
+  this.name = name;
+  this.health = 300;
```
nickborbe: If you do `this.health = health` it will pass the test as well because in the test, we are creating a new soldier and passing in 300 and 150 as arguments so that will ensure the values are set correctly.  The wording is the test was a little confusing.

nickborbe:
Nice work.

PR #28, opened 2017-02-20
MIA PT Colby Allen

colbyallenc:


starter-code/lib/viking.js
```diff
@@ -19,13 +19,82 @@
 //   FootViking: FootViking
```
nickborbe: You should keep this assignment to reference back to later.  Hope you understand what you did here b/c it's a good example of prototypal inheritance.

nickborbe:
Nice work.

PR #29, opened 2017-02-21
MIA PT Miguel Maldonado

migmaldonado:
Thanks!

### DIFF
--- a/starter-code/lib/viking.js
+++ b/starter-code/lib/viking.js
@@ -19,13 +19,83 @@
 //   FootViking: FootViking
 // };
 
-function Soldier () {}
+function Soldier () {
+  this.health   =300;
+  this.strength =150;
+  this.attack=function(){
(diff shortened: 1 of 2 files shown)


starter-code/lib/viking.js
```diff
@@ -19,13 +19,83 @@
...
+  };
+}
 
-function Viking () {}
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
 
-function Saxon () {}
+function Viking (name, health, strength) {
+  this.name     =name;
```
nickborbe: here, you can use `Soldier.call(this, health, strength)` to skip lines 37-39.  You can review the lecture on prototypal inheritance to see an example where we did it with `animal`

nickborbe:
Nice job.  Passing tests.

PR #30, opened 2017-02-22
MIA PT - Juan Duque

imth3walrus:


starter-code/lib/viking.js
```diff
@@ -19,13 +19,77 @@
...
+Saxon.prototype = Object.create(Soldier.prototype);
+
+Saxon.prototype.constructor = Saxon;
+
+function Saxon (name, health, strength) {
+  Soldier.call(this, health, strength);
+  this.health = 60;
+  this.strength = 25;
+  this.receiveDamage = function (damage) {
+    if (health >= this.health) {
```
nickborbe: `if (this.health <= 0`
makes more sense b/c this will check if the guy is alive or dead

nickborbe:
Nice work.  Passing all the tests.  Seems like you got a good idea of how inheritance works.

PR #33, opened 2017-02-23
MIA PT Dianelis Insua

dianelisinsua:


starter-code/lib/viking.js
```diff
@@ -19,13 +19,86 @@
 //   FootViking: FootViking
```
nickborbe: Good work.  Passing all the tests.  Thanks for submitting.

PR #34, opened 2017-02-23
MIA PT Aleksandra Boskovic Vikings - done

sandrabosk:


starter-code/test/tests.js
```diff
@@ -206,7 +206,6 @@ describe ('WAR', function () {
   it ('should show the result of the viking attack', function () {
     war.vikingAttack.should.be.a('Function');
     war.vikingAttack().should.equal('A Saxon has died in combat');
```
nickborbe: Haha, did you remove this to pass the test?

starter-code/lib/viking.js
```diff
@@ -19,13 +19,78 @@
 //   FootViking: FootViking
```
nickborbe: Good work.  It seems like you figured out how to pass all the test except the one you changed

PR #35, opened 2017-02-25
MIA PT Ernesto Quesada vikings

Ernesto-Quesada:


starter-code/lib/viking.js
```diff
@@ -19,13 +19,82 @@
 //   FootViking: FootViking
```
nickborbe: Nice work.  It seems like you have a good understanding of inheritance.

PR #43, opened 2017-04-07
BCN-Sergi-Garcia-Andrea-Martins

nolegendhere:


starter-code/lib/viking.js
```diff
@@ -19,13 +20,197 @@
...
 
-function Viking () {}
 
-function Saxon () {}
 
-function War () {}
+var result = new War();
 
 module.exports = {
   Soldier: Soldier,
```
tawebbcn: Good job, great way to represent the fight between armies!

PR #46, opened 2017-04-07
BCN-Pierre-Giuseppe

gfrr:


starter-code/lib/viking.js
```diff
@@ -19,13 +19,142 @@
...
+  } else {
+    this.saxonAttack();
+    this.vikingAttack();
+  }
+};
+
+
+War.prototype.showStats = function () {
+  return this.vikingsArmy.length >= this.saxonsArmy.length ? "Vikings have won the war of the century!" : "Saxons have won!";
+};
```
tawebbcn: Good job! Great way to solve the war part

PR #52, opened 2017-04-09
BCN-ISAK-EDUARD

isaklafleur:
Some stuff left to do.. Please point me in right direction regarding the 
War.prototype.vikingAttack and War.prototype.saxonAttack methods.. Or please share solution.

### DIFF
--- a/starter-code/lib/viking.js
+++ b/starter-code/lib/viking.js
@@ -1,31 +1,142 @@
-// //------------------------------------------------------
-// // SOLDIER
-// //------------------------------------------------------
-// function Soldier() {}
-//
-// //------------------------------------------------------
-// // GENERAL
-// //------------------------------------------------------
-// function General() {}
-//
-// //------------------------------------------------------
-// // FOOTVIKING
-// //------------------------------------------------------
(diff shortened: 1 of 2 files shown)


starter-code/lib/viking.js
```diff
@@ -1,35 +1,146 @@
...
+
+//////////////////////////////
+// SAXON - AREA
+//////////////////////////////
+
+function Saxon (health, strength) {
+  this.health = health;
+  this.strength = strength;
+  this.attack = () => this.strength
+  Saxon.prototype.receiveDamage = damage => {
```
zapatran: prototype outside

starter-code/lib/viking.js
```diff
@@ -1,35 +1,146 @@
...
+*/
+
+//////////////////////////////
+// SAXON - AREA
+//////////////////////////////
+
+function Saxon (health, strength) {
+  this.health = health;
+  this.strength = strength;
+  this.attack = () => this.strength
```
zapatran: Thats arrow function is es6 but is not pointing to a function

PR #88, opened 2017-06-16
bcn: Sara and Miguel

Tyrannogina:


starter-code/src/viking.js
```diff
@@ -1,11 +1,85 @@
...
+};
+War.prototype.showStatus = function(){
+  if (this.saxonArmy.length === 0) {
+    return "Vikings have won the war of the century!";
+  }
+  if (this.vikingArmy.length === 0) {
+    return "Saxons have fought for their lives and survive another day...";
+  }
+  return "Vikings and Saxons are still in the thick of battle.";
+};
```
tawebbcn: Good job!
Remember to use prototypes when declaring new methods

PR #94, opened 2017-06-16
bcn: Gina-Jorge

blackc0d3:


starter-code/src/viking.js
```diff
@@ -1,11 +1,110 @@
...
+
+var continueFighting = true;
+while (continueFighting) {
+  war.vikingAttack();
+  continueFighting = war.showStatus();
+  if (continueFighting) {
+    war.saxonAttack();
+    continueFighting = war.showStatus();
+  }
+}
```
tawebbcn: Good job, nice way to solve it.
Remember to use prototypes when declaring new methods

PR #95, opened 2017-06-16
BCN - Alex Fontanellas

alexfontanellas:


starter-code/src/viking.js
```diff
@@ -1,11 +1,127 @@
...
+  if(random<0.5){
+    startWar.vikingAttack();
+  }
+  else{
+    startWar.saxonAttack();
+  }
+  keepOn = startWar.showStatus();
+  times++;
+}
+console.log("The winners of the fight are the", keepOn, "having fought", times, "times");
```
tawebbcn: Great job! 
Good use of prototypes and nice way to solve the war method

PR #97, opened 2017-06-16
BCN-christiana-julius

jsur:
done
@chrikoudi

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,119 @@
 // Soldier
-function Soldier () {}
+function Soldier (health, strength) {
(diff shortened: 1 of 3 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,119 @@
...
+  }
+  if (this['vikingArmy'].length > 0 && this['saxonArmy'].length) {
+    return "Vikings and Saxons are still in the thick of battle.";
+  }
+
+}
+
+function selectRandomIndex(array) {
+  return Math.floor(Math.random()* array.length);
+};
```
tawebbcn: good job!

PR #98, opened 2017-06-16
BCN - Maria Mayer & Laura

lauravallhonrat:
done

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,70 @@
-// Soldier
-function Soldier () {}
+//iteration 1 soldier
+function Soldier(health,strength){
+  this.health = health;
+  this.strength = strength;
+}
 
-// Viking
-function Viking () {}
+Soldier.prototype.attack = function(){
+    return this.strength;
+}
+  
+Soldier.prototype.receiveDamage = function(damage){
+    return this.health -= damage;
+}
+  
+var soldier = new Soldier(300,150);
 
-// Saxon
-function Saxon () {}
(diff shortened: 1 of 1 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,70 @@
...
+  Soldier.prototype.receiveDamage.call(this,damage);
+  if(this.health <= 0){
+    return "A saxon has died in act of combat";
+    
+  }else{
+    return "A saxon has received "+ damage +" points of damage";
+  }
+}
+var saxon = new Saxon(60,25);
+saxon.receiveDamage(10);
```
tawebbcn: Good job and good use of prototypes, try to go for the bonus part if you have the time :)

starter-code/src/viking.js
```diff
@@ -1,11 +1,70 @@
...
+function Viking(name,health,strength){
+  Soldier.call(this,health,strength);
+  this.name = name;
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
+
+Viking.prototype.receiveDamage = function (damage){
+  
+  Soldier.prototype.receiveDamage.call(this,damage);
```
matilore: BAD! You are already inheriting that method from the soldier constructor, having two methods with the same name among inheriting "classes" means that you should use only one for all classes, and not trying to use one inside the other belonging to the other class

matilore:
The code is good but you are not using the parent constructor/"class" soldier in the proper way, let me know if you have doubts

PR #99, opened 2017-06-16
Camille (and JB a bit...)

camdw:


matilore:
The prototypal inheritance code is correct but you are using it in the proper way, ask me later if you have doubts

PR #100, opened 2017-06-16
BCN - Jonathyn (JB) Tellez and Luis Amez 

JB-Tellez:
Vikings Pair Programming code

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,161 @@
 // Soldier
-function Soldier () {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+}
+
+Soldier.prototype.attack = function () {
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,165 @@
...
+    Soldier.call(this, health, strength);
+    this.name = name;
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
+
+Viking.prototype.receiveDamage = function (damage) {
+
+    // TA says stupid JavaScript cannot call super method like below, so just duplicate the code
+    // Soldier.prototype.receiveDamage.call(this, damage);
```
matilore: Why would you do something like that if you are already inheriting vikings from soldier? you already have that method declared as a viking method

matilore:

PR #107, opened 2017-06-18
All but attack methods - still need to finalize

bradford-le:
Attended review session so worked on this by myself with help of other IHackers
Question: For my War constructor, I initialized a war at the beginning so in console I can add my viking that I created. Is this correct?


matilore:
The code is good but you are duplicating code among saxons and vikings and the main reason is that you are not using the parent constructor/"class"  soldier in the proper way, let me know if you have doubts

PR #108, opened 2017-06-18
Alex Buchan Vikings Js

alexbuchan:


matilore:
The code is good but I would prefer you to use prototypal inheritance, you are duplicating code among saxons and vikings and the main reason is that you are not using the parent constructor/"class" in the proper way

PR #110, opened 2017-06-18
BCN Cillian

Chilliano:
Correcto


matilore:
I can see you have to complete it but so far so very good but try to use prototypal inheritance! ;)

PR #125, opened 2017-08-11
MIA Joao & Ryan

RyanCommits:


dsoraluz:
Great job applying what we have learned up to this point. There's a couple of different ways that you could have approached this and it seems that you guys took the time to think it through and keep it to best practices. 

PR #127, opened 2017-08-11
MIA Danny and Ernesto, Assigment completed someone won the war

shebbbby:


starter-code/src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+    this.vikingArmy.splice(randomVikingNumber);
+  }
+  return returnedV;
+};
+
+War.prototype.showStatus = function(){
+  if(this.vikingArmy.length > 0 && this.saxonArmy.length > 0){
+    return "Vikings and Saxons are still in the thick of battle."
+  }
+  if(this.saxonArmy.length === 0){
```
dsoraluz: Good job with the logic here. Multiple if statements definitely get the job done but next time maybe consider how other decision structures can be used (the one I'm thinking of here is the if...else if... else structure).

dsoraluz:
Good job applying what we have learned up to this point. This was a good example to help you really drive home the use of prototypes and some helpful logic that can be applied to your upcoming games for the end of module 1.

PR #128, opened 2017-08-11
MIA Dharma and Jamilet

DharmaG:


dsoraluz:
Great job completing most of the exercise. The war logic can be a little tricky because it is complex but the best way to take on a complex problem is to divide it into small pieces and attempt to solve them one by one.

I would have liked to seen you guys attempt the war because that would have given you a good idea of how game logic works and that would have given you a good example of some helpful logic that can be applied to your upcoming games for the end of module 1.

PR #137, opened 2017-08-18
BCN - Mark, Tania and Markus

markvertra:
@MHFP
@taniabatista

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,134 @@
 // Soldier
-function Soldier () {}
+function Soldier (health, strength) {
+	this.health = health;
+	this.strength = strength;
+	
+}
+
+Soldier.prototype.attack = function(){
+		return this.strength;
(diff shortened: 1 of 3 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,134 @@
...
+War.prototype.addSaxon = function(saxon){
+	this.saxonArmy.push(saxon);
+}
+
+War.prototype.vikingAttack = function(){
+    if (this.saxonArmy.length === 0) {
+		return "Saxons have fought for their lives and survive another day...";
+	}
+	// select random Saxon from array
+	saxon = this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
```
tawebbcn: You are doing this four times. Maybe you can use a function.

tawebbcn:
Be careful with the indentation.

Great job! :+1:

PR #142, opened 2017-08-21
BCN - Frances & Jordi

frances0688:
@jviles


tawebbcn:
Good job, but try the bonus if you have time. :+1:

PR #151, opened 2017-09-08
PAR - Reshma & Valentin - vikings

kira912:


starter-code/src/viking.js
```diff
@@ -1,11 +1,109 @@
...
+  }
+  return hit;
+};
+
+function randomSelect (array) {
+  var index = Math.floor(Math.random() * array.length);
+
+  return {
+    item: array[index],
+    index: index
```
yacinehmito: Yes! I wanted to show it to the students but I couldn't; I'm glad you did it this way.

PR #169, opened 2017-10-20
BCN - Jean and Dominik

dlasch089:
@jnothomb


starter-code/src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+  this.vikingArmy.push(viking);
+};
+
+War.prototype.addSaxon = function(saxon) {
+  this.saxonArmy.push(saxon);
+};
+
+War.prototype.vikingAttack = function() {
+  var randomSaxon = Math.floor(Math.random()*this.saxonArmy.length);
+  var randomViking = Math.floor(Math.random()*this.vikingArmy.length);
```
tawebbcn: You are repeating the same code. Maybe you can use a function for selecting a random soldier...

starter-code/src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+  var saxonNewHealth  = this.saxonArmy[randomSaxon].receiveDamage(this.vikingArmy[randomViking].strength);
+  if (this.saxonArmy[randomSaxon].health <= 0){
+    this.saxonArmy.pop(this.saxonArmy[randomSaxon]);
```
tawebbcn: The pop method doesn't accept an argument, always remove the LAST element of the array. If you want to remove one specifc element in an array... ¡¡¡SEARCH IN GOOGLE!!!

Ok, here you have a hint, but don't get used: https://stackoverflow.com/questions/5767325/how-do-i-remove-a-particular-element-from-an-array-in-javascript

starter-code/src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+  }
+return saxonNewHealth;
+};
+
+War.prototype.saxonAttack = function() {
+  var randomSaxon = Math.floor(Math.random()*this.saxonArmy.length);
+  var randomViking = Math.floor(Math.random()*this.vikingArmy.length);
+  var vikingNewHealth  = this.vikingArmy[randomViking].receiveDamage(this.saxonArmy[randomSaxon].strength);
+  if (this.vikingArmy[randomViking].health <= 0){
+    this.vikingArmy.pop(this.vikingArmy[randomViking]);
```
tawebbcn: Same as in vikingAttack().

PR #170, opened 2017-10-20
BCN - Lucas Affini / Clara Ameller

lacazeto:
@claraameller

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,98 @@
 // Soldier
-function Soldier () {}
+function Soldier (health, strength) {
+    this.health = health;
+    this.strength = strength;
+
+    this.attack = function(){
+        return this.strength;
+    };
+
+    this.receiveDamage = function(damage) {
+        this.health -= damage;
+    }
+}
 
 // Viking
-function Viking () {}
+function Viking (name, health, strength) {
+    Soldier.call(this, health, strength);
+    this.name = name;
+
+    this.receiveDamage = function(damage) {
+        this.health -= damage;
+        if(health > 0){
+            return this.name + " has received " + this.damage + " points of damage";
+        }
+        else {
+            return this.name + " has died in act of combat";
+        }
+    }
+
+    this.battleCry = function(){
+        return "Odin Owns You All!";
+    }
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
 
 // Saxon
-function Saxon () {}
+function Saxon (health, strength) {
+    Soldier.call(this, health, strength);
+
+    this.receiveDamage = function(damage) {
+        this.health -= damage;
+        if(health > 0){
+            return "A Saxon has received " + this.damage + " points of damage";
+        }
+        else {
+            return "A Saxon has died in act of combat";
+        }
+    }
+}
+
+Saxon.prototype = Object.create(Soldier.prototype);
 
 // War
-function War () {}
+function War () {
+    this.vikingArmy = [];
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
 // Soldier
-function Soldier () {}
+function Soldier (health, strength) {
+    this.health = health;
+    this.strength = strength;
+
+    this.attack = function(){
```
tawebbcn: Don't define the object's methods inside the constructor. Instead, outside it, do:
```
Soldier.prototype.attack = function () {
  return this.strength;
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+        return this.strength;
+    };
+
+    this.receiveDamage = function(damage) {
```
tawebbcn: Same as before:
```
Soldier.prototype.receiveDamage = function (damage) {
  this.health -= damage;
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+        this.health -= damage;
+    }
+}
 
 // Viking
-function Viking () {}
+function Viking (name, health, strength) {
```
tawebbcn: Same problem, methods better outside the constructor.

starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+            return this.name + " has died in act of combat";
+        }
+    }
+
+    this.battleCry = function(){
+        return "Odin Owns You All!";
+    }
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
```
tawebbcn: After that, for proper inheritance's implementation, you need to add:
```
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
 
 // Saxon
-function Saxon () {}
+function Saxon (health, strength) {
```
tawebbcn: Same problems as in Viking.

starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+            return "A Saxon has died in act of combat";
+        }
+    }
+}
+
+Saxon.prototype = Object.create(Soldier.prototype);
 
 // War
-function War () {}
+function War () {
```
tawebbcn: One more time, methods outside the constructor.

starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+        var theVictm = Math.floor(Math.random() * saxonArmy.length); 
+        var isDead = saxonArmy[theVictm].receiveDamage(300);
+        if (isDead = "A Saxon has died in act of combat"){
+            this.saxonArmy.splice(theVictm,1);
+        }
+
+        return isDead;
+    }
+
+    this.saxonAttack = function(){
```
tawebbcn: This code looks exactly the same as vikingAttack(). Maybe you can use a function...

starter-code/src/viking.js
lacazeto: Why is that? 
Wouldn´t this make more susceptible to error-prone?

tawebbcn: Mainly for performance and economy. We talked about this when you were doing the exercise. Of course, we can discuss it again. Anyways, here you have more info: https://stackoverflow.com/questions/4508313/advantages-of-using-prototype-vs-defining-methods-straight-in-the-constructor

PR #172, opened 2017-10-20
BCN - Bryan, Cristian y Sara M.

saringa:
@cristiandiazz
@brianboyCode


tawebbcn:
Great job with object definition and inheritance. Keep the hard work! :+1:

PR #174, opened 2017-10-20
BCN - PP Vikings _ Desiree - Elvin

3lv27:
@dvillool


tawebbcn:
Good job with object definition and inheritance. Keep the hard work! :+1:

PR #175, opened 2017-10-20
BCN Misha - Jordi

joordiii:
@MythicalMichael

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,69 @@
 // Soldier
-function Soldier () {}
+function Soldier (health,strength) {
+this.health=health;
+this.strength=strength;
+this.attack=function(){
+  return this.strength;
+};
+this.receiveDamage=function(damage){
+this.health -= damage;
+};
+
+}
 
 // Viking
-function Viking () {}
+function Viking (name,health,strength) {
+Soldier.call(this,health,strength);
+this.name=name;
+this.attack=function(){
+return this.strength;
+};
+this.receiveDamage=function(damage){
+this.health -= damage;
+if (this.health>0){
+  return (this.name + ' has received ' + damage + ' points of damage');
+}
+else {
+  return this.name +' has died in act of combat';
+}
+};
+this.battleCry=function(){
+return 'Odin Owns You All!';
+};
+}
+Viking.prototype=Object.create(Soldier.prototype);
 
 // Saxon
-function Saxon () {}
+function Saxon (health,strength) {
+Soldier.call(this,health,strength);
+this.attack=function(){
+return (this.attack=this.strength);
+};
+this.receiveDamage=function(damage){
+this.health -= damage;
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
 // Soldier
-function Soldier () {}
+function Soldier (health,strength) {
+this.health=health;
```
tawebbcn: What about the indentation?

starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
...
+this.strength=strength;
+this.attack=function(){
```
tawebbcn: Don't define the object's methods inside the constructor. Instead, outside it, do:
```
Soldier.prototype.attack = function () {
  return this.strength;
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
...
+  return this.strength;
+};
+this.receiveDamage=function(damage){
```
tawebbcn: Same as before:
```
Soldier.prototype.receiveDamage = function (damage) {
  this.health -= damage;
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
...
+this.health -= damage;
+};
+
+}
 
 // Viking
-function Viking () {}
+function Viking (name,health,strength) {
```
tawebbcn: Same problem, methods better outside the constructor.

starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
...
+}
+else {
+  return this.name +' has died in act of combat';
+}
+};
+this.battleCry=function(){
+return 'Odin Owns You All!';
+};
+}
+Viking.prototype=Object.create(Soldier.prototype);
```
tawebbcn: After that, for proper inheritance's implementation, you need to add:
```
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
...
 
 // Saxon
-function Saxon () {}
+function Saxon (health,strength) {
```
tawebbcn: Same problems as in Viking.

starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
...
+Soldier.call(this,health,strength);
+this.attack=function(){
+return (this.attack=this.strength);
```
tawebbcn: You want to return this.strength, not this.attack=this.strength.

starter-code/src/viking.js
```diff
@@ -1,11 +1,69 @@
...
+}
+};
+
+}
+Saxon.prototype=Object.create(Soldier.prototype);
+
 
 // War
-function War () {}
+function War () {
```
tawebbcn: One more time, methods outside the constructor.

PR #178, opened 2017-10-20
- BCN IronHack Fco.David feat Dafne

otidlam:
@dafneolca


tawebbcn:
Nice start and good implementation of inheritance. :+1:

PR #184, opened 2017-10-23
MIA - Ian Izaguirre - Marcos Palacios

ianizaguirre:
Vikings


dsoraluz:
Good job applying what we have learned up to this point. This was a good example to help you really drive home the use of prototypes and some helpful logic that can be applied to your upcoming games for the end of module 1.

PR #230, opened 2018-01-12
PAR viking saxon v1 - Maureen & Rithy

Bentchaw:


starter-code/src/viking.js
```diff
@@ -1,11 +1,71 @@
...
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+  this.name = name;
+
+  Soldier.call(this, health, strength);
+
+  this.receiveDamage = function(damage) {
```
ta-web-paris: remember to add this function to the prototype of Viking instead just after line 34

starter-code/src/viking.js
```diff
@@ -1,11 +1,71 @@
...
+    this.health -= damage;
+    if (this.health > 0) {
+      return this.name + " has received" + damage + " points of damage";
+    } else {
+      return this.name + " has died in act of combat";
+    }
+  };
+
+  function battleCry() {
```
ta-web-paris: This function has to be added to the prototype too

ta-web-paris:
Good job but make sure to check the correction for the functions being added to the prototype

PR #250, opened 2018-01-15
BCN - Giovanni and Daniel

JoniWanKenobi:
@DTDK

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,72 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+    this.attack = function(){
+        return this.strength;
+    };
+    this.receiveDamage = function(damage){
+        this.health -= damage;
+    };
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    Soldier.call(this, health, strength);
+    this.name = name;
+    this.receiveDamage = function(damage){
+        this.health -= damage;
+        if(this.health>0){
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,72 @@
...
+        }
+    };
+    this.battleCry = function(){
+        return "Odin Owns You All!"
+    }
+
+};
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
```
tawebbcn: If you use the "simpler" version of oop, you don't need this two lines (32 and 33). If you add your methods with prototype, then you will need them.

starter-code/src/viking.js
```diff
@@ -1,11 +1,72 @@
...
+            return "A Saxon has received " + damage + " points of damage";
+        }
+        else {
+            return "A Saxon has died in combat";
+        }
+    };
+}
+
+Saxon.prototype = Object.create(Soldier.prototype);
+Saxon.prototype.constructor = Saxon;
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,72 @@
...
+        this.vikingArmy.push(viking);
+
+    };
+    this.addSaxon = function(saxon){
+        this.saxonArmy.push(saxon);
+    };
+    this.vikingAttack = function(){
+        var vIndex = Math.floor(Math.random()*this.vikingArmy.length);
+        var sIndex = Math.floor(Math.random()*this.saxonArmy.length);
+        var res = this.saxonArmy[sIndex].receiveDamage(this.vikingArmy[vIndex].strength);
```

PR #251, opened 2018-01-15
BCN - vikings vlad and oscar

vladcio:
@oscarjorgejorge

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,105 @@
 // Soldier
-function Soldier() {}
+
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+}
+
+
+Soldier.prototype.attack = function () {
+    return this.strength;
+}
+
+Soldier.prototype.receiveDamage = function (damage) {
+    this.health = this.health - damage;
+}
+
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    Soldier.call(this, health,strength);
+    this.name = name;
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+}
 
 // War
-function War() {}
+function War() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+}
+
+War.prototype.addViking = function(Viking){
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+    this.vikingArmy.push(Viking) 
+}
+
+War.prototype.addSaxon = function(Saxon){
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+    this.saxonArmy.push(Saxon) 
+}
+
+War.prototype.vikingAttack = function () {
+    var randomViking = Math.floor(Math.random() * this.vikingArmy.length);
+    var randomSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+
+    var warResult = this.saxonArmy[randomSaxon].receiveDamage(this.vikingArmy[randomViking].strength);
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+    }
+
+    return warResult;
+}
+
+War.prototype.saxonAttack = function () {
+    var randomViking = Math.floor(Math.random() * this.vikingArmy.length);
+    var randomSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+
+    var warResult = this.vikingArmy[randomViking].receiveDamage(this.saxonArmy[randomSaxon].strength);
```

PR #254, opened 2018-01-15
BCN Stephie & Sheng

kissy89:
@therealsheng

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -53,4 +53,66 @@ Saxon.prototype.receiveDamage = function(damage){
 }
 
 // War
-function War() {}
+function War() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+}
+
+
+War.prototype.addViking = function(viking){
+    this.vikingArmy.push(viking);
+}
+
+War.prototype.addSaxon = function(saxon){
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,118 @@
...
+
+War.prototype.addSaxon = function(saxon){
+    this.saxonArmy.push(saxon);
+}
+
+War.prototype.vikingAttack = function(){
+    var saxon = this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
+    var viking = this.vikingArmy[Math.floor(Math.random()*this.vikingArmy.length)];
+
+    if (saxon.health < viking.strength) {
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,118 @@
...
+//     return saxon.receiveDamage(viking.strength);
+// }
+
+War.prototype.saxonAttack = function(){
+    var saxon = this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
+    var viking = this.vikingArmy[Math.floor(Math.random()*this.vikingArmy.length)];
+
+    var result = viking.receiveDamage(saxon.attack());
+
+    if (viking.health < saxon.strength) {
```

PR #255, opened 2018-01-15
BCN - Danielle & Linda

danielleleb:
@lindaluisa


starter-code/src/viking.js
```diff
@@ -1,11 +1,95 @@
...
 // Viking
-function Viking() {}
+function Viking(name, healthArg, strengthArg) {
+
+    Soldier.call(this, healthArg, strengthArg)
+    this.name = name;
+    this.receiveDamage = function(damage) {
+        this.health -= damage;
+        if (this.health > 0){
+        return this.name + " has received " + damage + " points of damage";
```
tawebbcn: Be careful with indentation.

starter-code/src/viking.js
```diff
@@ -1,11 +1,95 @@
...
+            return this.name + " has died in act of combat";
+        }
+    }
+
+    this.battleCry = function() {
+        return "Odin Owns You All!";
+    }
+}
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
```
tawebbcn: If you use the "simpler" version of oop, you don't need this two lines (31 and 32). If you add your methods with prototype, then you will need them.

starter-code/src/viking.js
```diff
@@ -1,11 +1,95 @@
...
+    this.receiveDamage = function(damage) {
+        this.health -= damage;
+        if (this.health > 0){
+        return "A Saxon has received " + damage + " points of damage";
+        } else {
+            return "A Saxon has died in combat";
+        }
+    }
+}
+Saxon.prototype = Object.create(Soldier.prototype);
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,95 @@
...
 
 // War
-function War() {}
+function War() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+    this.addViking = function(Viking){
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,95 @@
...
+        this.vikingArmy.push(Viking);
+    } 
+    this.addSaxon = function(Saxon) {
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,95 @@
...
+        this.saxonArmy.push(Saxon);
+    } 
+    this.vikingAttack = function() {
+        this.randomSaxon = this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)];
+        this.randomViking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
+
+        var result = this.randomSaxon.receiveDamage(this.randomViking.strength);
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,95 @@
...
+
+        return result;
+       
+    }
+
+    this.saxonAttack = function () {
+        this.randomSaxon = this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)];
+        this.randomViking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
+
+        var result = this.randomViking.receiveDamage(this.randomSaxon.strength);
```

PR #256, opened 2018-01-15
BCN - Pedro & Alla

allarom:
@pedrodiogoneto

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,75 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+}
+
+Soldier.prototype.attack = function () {
+    return this.strength;
+};
+
+Soldier.prototype.receiveDamage = function (damage) {
+    this.health -= damage;
+};
+
 
 // Viking
-function Viking() {}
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,75 @@
...
+};
 
 // War
-function War() {}
+function War() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+}
+
+War.prototype.addViking = function (Viking){
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,75 @@
...
+    this.vikingArmy.push(Viking);
+}
+
+War.prototype.addSaxon = function (Saxon){
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,75 @@
...
+    this.saxonArmy.push(Saxon);
+}
+
+
+War.prototype.vikingAttack = function() {
+    var s1 = this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
+    var v1 = this.vikingArmy[Math.floor(Math.random()*this.vikingArmy.length)];
+    s1.receiveDamage(v1.attack());
```
tawebbcn: You missed checking the saxon health and return the result of receiveDamage().

PR #258, opened 2018-01-15
BCN - David & Victor

davidalexteixeira:
@victor.pellerin

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -87,5 +87,12 @@ War.prototype.addSaxon = function (saxon) {
 War.prototype.vikingAttack = function () {
    var randomViking = this.vikingArmy[Math.floor(Math.random()*this.vikingArmy.length)];
    var randomSaxon = this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
-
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+    this.saxonArmy.push(saxon);
+
+}
+
+
+War.prototype.vikingAttack = function () {
+   var randomViking = this.vikingArmy[Math.floor(Math.random()*this.vikingArmy.length)];
+   var randomSaxon = this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
+   var index = this.saxonArmy.indexOf(randomSaxon)
+    randomSaxon.receiveDamage(randomViking.strength)
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+   
+    if(randomSaxon.health <= 0){
+     this.saxonArmy.splice(index, 1);
+     return 'A Saxon has died in combat';
```
tawebbcn: Instead of returning that string here (then you miss the return when the saxon survives), store the result of receiveDamage() in a variable and return that at the end (line 97).

PR #259, opened 2018-01-15
BCN - Santiago and Rodrigo

santironhacker:
@rodrigodaoud

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,81 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;    
+    this.attack = function() {
+        return strength;
+    }
+}
+
+//     receiveDamage();    
+Soldier.prototype.receiveDamage = function(damage) {
+    this.health -= damage;
+}
+   
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    this.name = name;
+    Soldier.call(this, health, strength);
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
+
+Viking.prototype.receiveDamage = function(damage) {
+    this.health -= damage;
+    if(this.health > 0) {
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,81 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;    
+    this.attack = function() {
```
tawebbcn: Try to be consistent. Except for this one, you are declaring the methods for Soldier, Viking and Saxon outside the constructor (recommended). Put this one outside too.

starter-code/src/viking.js
```diff
@@ -1,11 +1,81 @@
...
+    }
+}
+
 
 // War
-function War() {}
+function War() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+    this.addViking = function(object) {
```
tawebbcn: This code is correct, but better call the parameter viking, rather than object. Futhermore, once again, be consistent. Put the methods outside the constructor.

starter-code/src/viking.js
```diff
@@ -1,11 +1,81 @@
...
+        this.vikingArmy.push(object);
+    }
+    this.addSaxon = function(object) {
```
tawebbcn: Same as before, but with saxon.

starter-code/src/viking.js
```diff
@@ -1,11 +1,81 @@
...
+        this.saxonArmy.push(object);
+    }
+    this.vikingAttack = function () {
+        var vIndex = Math.floor(Math.random()*this.vikingArmy.length);
+        var sIndex = Math.floor(Math.random()*this.saxonArmy.length);
+        var hp = this.saxonArmy[sIndex].receiveDamage(this.vikingArmy[vIndex]);
```
tawebbcn: You are missing .attack(). receiveDamage is expecting a number and you are passing a viking.
```
var hp = this.saxonArmy[sIndex].receiveDamage(this.vikingArmy[vIndex].attack());
```

PR #265, opened 2018-01-26
PP MEX Alicia Dimitri

Dimbnnr:
- Complicado. . .
- Nos tardamos en entender los metodos  y la relación con Jasmin etc..
- Creo que aun no lo entiendo (Dimitri)

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,74 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+this.health = health;
+this.strength = strength
+}
+
+Soldier.prototype.attack = function(){
+    return this.strength;
+}
+
+Soldier.prototype.receiveDamage = function(damage){
+     this.health = this.health - damage;
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,74 @@
...
+
+War.prototype.addViking = function(Viking){
+    this.vikingArmy.push(Viking);
+}
+
+War.prototype.addSaxon = function(Saxon){
+    this.saxonArmy.push(Saxon);
+}
+
+War.prototype.vikingAttack = function(){
```
ta-web-mex: Una posible solucion:

    this.vikingAttack = function(){
        var randomSaxon = Math.floor(Math.random()*this.saxonArmy.length);
        var randomViking = Math.floor(Math.random()*this.vikingArmy.length);
        this.saxonArmy[randomSaxon].receiveDamage(this.vikingArmy[randomViking].strength);
        if (this.saxonArmy[randomSaxon].health<=0) {
            this.saxonArmy.splice(randomSaxon,1);
            return "A Saxon has died in combat";
        } else {
            return "A Saxon has received " + this.vikingArmy[randomViking].strength + " points of damage";
        }
    };

PR #268, opened 2018-01-27
[MIA] [Stephen - Jackson]

stephjackson:
this lesson explains oop, how to do it, and why it's important better than the college class i took on it.


sandrabosk:
Very good. The only thing I could recommend is to organize the code more functionally, in blocks of code related to Vikings, Saxons and War. But that is really minor thing.
Thank you for submitting.

PR #269, opened 2018-01-28
[MIA][Mosi Rivera]

mosileko:


starter-code/src/viking.js
```diff
@@ -1,11 +1,79 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+
+  this.attack = function () {
```
sandrabosk: The best practice is to use methods on prototypes so here should be something like:
```
  return this.strength;
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,79 @@
...
+    return this.strength;
+  };
+
+  this.receiveDamage = (damage) => {
```
sandrabosk: The same as above ^

starter-code/src/viking.js
```diff
@@ -1,11 +1,79 @@
...
+  };
+}
 
 // Viking
-function Viking() {}
+
+function Viking(name, health, strength) {
+  Soldier.call(this, health, strength);
+  this.name = name;
+  this.receiveDamage = (damage) => {
```
sandrabosk: All methods should be defined on prototypes.

starter-code/src/viking.js
```diff
@@ -1,11 +1,79 @@
...
+    this.health -= damage;
+    return this.health > 0 ? this.name + ' has received ' + damage + ' points of damage' : this.name + ' has died in act of combat';
+  };
+  this.battleCry = function () {
+    return 'Odin Owns You All!';
+  };
+}
+Viking.prototype = Object.create(Soldier.prototype);
```
```

```
So from now on we could re-implement method receivedDamage from Soldier:
```
Viking.prototype.receiveDamage = function(damage) { 
         ....
};
```
and battleCry:
```
  return "Odin Owns You All!";
};
```

sandrabosk:
Good understanding, but check the comments to give you better perspective of using prototype concepts. Thank you for submitting.

PR #270, opened 2018-01-28
[MIA][Scott-Rowell]

rowellscott:


sandrabosk:
Very good. It seems you're master of prototypal inheritance. ;-)
Thank you for submitting.

PR #271, opened 2018-01-29
[MIA] [Nicolas - Levo]

nlevo:


sandrabosk:
Very good. It looks like you're master of prototypal inheritance. Thank you for submitting.

PR #272, opened 2018-01-29
[MIA] [GYASIK - NUNEZ]

gyasikn:


starter-code/src/viking.js
```diff
@@ -1,11 +1,92 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+  this.attack = function () {
```
```
  return this.strength;
};
```

sandrabosk:
You understood basic concepts, just keep an eye on syntax and make sure you use inheritance since it's very common practice. Thank you for submitting.

PR #274, opened 2018-01-29
[MIA] Martin-Genet

CortezExpress:


starter-code/src/viking.js
```diff
@@ -1,11 +1,72 @@
 // Soldier
-function Soldier() {}
+function Soldier (health, strength) {
+    this.health = health;
+    this.strength = strength;
+    this.attack = function(){
```
```
  return this.strength;
};
```

sandrabosk:
So far good enough. If you find time, keep working on this exercise since it will help you to understand basic concepts of inheritance. Thank you for submitting.

PR #276, opened 2018-01-30
[MIA][Francis-Calizo]

FrancisCalizo:
Everything besides bonus


starter-code/src/viking.js
```diff
@@ -1,11 +1,63 @@
 // Soldier
-function Soldier() {}
+//Constructor function receving 2 arguments
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+  
+  this.attack = function(){
```
```
  return this.strength;
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+//Viking
+function Viking(name, health, strength) {
+  //inheriting from Soldier constructor
+  Soldier.call(this, health, strength);
+  //property unique to viking
+  this.name = name;
 
-// War
-function War() {}
+  this.receiveDamage = function(damage){
```
sandrabosk: The same as before, we should be using this syntax:
```
Viking.prototype.receiveDamage = function(damage) { ... }
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+    this.health -= damage;
+    if(this.health > 0){
+      //this keyword to reference the viking function
+      return this.name + " has received " + damage + " points of damage";
+    } else{
+      return this.name + " has died in act of combat"
+    }
+  }
+
+  this.battleCry = function() {
```
```
  return "Odin Owns You All!";
};
```

sandrabosk:
You understand good enough but keep an eye on syntax we use in inheritance. Thank you for submitting.

PR #277, opened 2018-01-30
[MIA] [Danny-Autrey]

Dautrey94:


starter-code/src/viking.js
```diff
@@ -1,9 +1,32 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
 
+    Soldier.prototype.attack = function(){
+    return this.strength();
```
```
this.strength
```
is a property of the Soldier, not a function, so no ().

PR #279, opened 2018-01-30
[Mia] [Michael - Elso]

ElsoMichael:
Didn't know how to tackle the bonus


sandrabosk:
Very good approach. It seems you understood the basics of prototypal inheritance. Thank you for submitting.

PR #282, opened 2018-01-30
[MIA] Rakeem Thomas JavaScript Vikings

rthomas29:
I enjoyed completing this exercise! Also, I like how we're slowly being exposed to Jasmine and creating assertions, piece by piece.


starter-code/src/viking.js
```diff
@@ -1,11 +1,91 @@
...
 
 // War
-function War() {}
+function War() {
+  this.vikingArmy = [];
+  this.saxonArmy = [];
+}
+
+War.prototype.addViking = function(Viking) {
+  this.vikingArmy.push(Viking);
```
sandrabosk: The common practice is to use capital letters for constructor functions and small letters for parameters.

PR #300, opened 2018-03-23
BCN samantha - miriam!

samanthaferreira:
@MiriamAparicio

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,57 @@
-// Soldier
-function Soldier() {}
+// Soldier constructor
+function AbstractSoldier(healthArg, strengthArg) {
+    this.health = healthArg;
+    this.strength = strengthArg;
+}
+Soldier.prototype.attack = function () {
+    return this.strength;
+}
 
-// Viking
-function Viking() {}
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,57 @@
-// Soldier
-function Soldier() {}
+// Soldier constructor
+function AbstractSoldier(healthArg, strengthArg) {
+    this.health = healthArg;
+    this.strength = strengthArg;
+}
+Soldier.prototype.attack = function () {
```
tawebbcn: Careful here, you changed the name of the constructor so every time you used Soldier, you needed to use AbstractSoldier

starter-code/src/viking.js
```diff
@@ -1,11 +1,57 @@
...
+
+
+// Viking constructor 
+function Viking(name, health, strength) {
+    Soldier.call(this, health, strength);
+    this.name = name;
+}
+
+Viking.prototype.attack = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
```
tawebbcn: Good :+1:

PR #302, opened 2018-03-23
BCN-MARKETA-GERMAN

GermanRodrickson:
@Hrabenka7

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,54 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+}
+Soldier.prototype.attack= function() {
+  return this.strength;
+}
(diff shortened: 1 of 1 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,54 @@
...
+
+Viking.prototype.battelCry = function() {
+  return 'Odin Owns You All'
+};
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
+
+Viking.prototype.recibeDamage = function (damage) {
+  Soldier.prototype.recibeDamage.call(this, damage);
```
tawebbcn: No need to do this. The methods are inherited as well!

starter-code/src/viking.js
```diff
@@ -1,11 +1,54 @@
...
+ // this.health = this.health - damage;
```
tawebbcn: This was good!

PR #304, opened 2018-03-23
BCN - Sebastian - Ruben

Sebachowa:
@rubenaguayo

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,92 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+};
+
+Soldier.prototype.attack = function() {
+  return this.strength;
+};
+
+Soldier.prototype.receiveDamage = function(damage) {
+  this.health -= damage;
+};
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,104 @@
...
+    this.saxonArmy.splice(this.vikingArmy.indexOf(saxon), 1);
+  }; 
+  return result;
+};
+
+War.prototype.saxonAttack = function() {
+  var saxonDamage = selectRandomSoldier(this.saxonArmy).strength;
+  var viking = selectRandomSoldier(this.vikingArmy);
+  var result = viking.receiveDamage(saxonDamage);
+  if (viking.health < 1) {
```
tawebbcn: 👍

starter-code/src/viking.js
```diff
@@ -1,11 +1,104 @@
...
+};
+
+War.prototype.showStatus = function() {
+  if (this.saxonArmy.length === 0) {
+    return "Vikings have won the war of the century!";
+  } else if (this.vikingArmy.length === 0) {
+    return "Saxons have fought for their lives and survive another day...";
+  } else {
+    return "Vikings and Saxons are still in the thick of battle.";
+  };
```
tawebbcn: Sice you have `return` you don't need the `else`. Only `if`

PR #306, opened 2018-03-23
bcn-dauv-fredrik

fredrik62:
@dauvbafna

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,70 @@
 // Soldier
-function Soldier() {}
+function Soldier(healthArg, strengthArg) {
+    this.health = healthArg;
+    this.strength = strengthArg;
+}
+
+Soldier.prototype.attack = function() {
+    return this.strengthArg;
+}
+
+Soldier.prototype.receiveDamage = function(damage) {
+   return Soldier.healthArg - damage;
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    Soldier.call(this, healthArg, strengthArg);
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
+
+
+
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,70 @@
 // Soldier
-function Soldier() {}
+function Soldier(healthArg, strengthArg) {
+    this.health = healthArg;
+    this.strength = strengthArg;
+}
+
+Soldier.prototype.attack = function() {
+    return this.strengthArg;
```
tawebbcn: `this.strengthArg`doesn't exist on the `Soldier` constructor. What you have is `this.health` and `this.strength`

starter-code/src/viking.js
```diff
@@ -1,11 +1,70 @@
...
+}
+
+Soldier.prototype.receiveDamage = function(damage) {
+   return Soldier.healthArg - damage;
```
tawebbcn: `Soldier.healthArg` is nothing. You should have something like:

```
return this.health -= this.damage
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,70 @@
...
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    Soldier.call(this, healthArg, strengthArg);
```
tawebbcn: `healthArg, strengthArg` doesn't exist here

starter-code/src/viking.js
```diff
@@ -1,11 +1,70 @@
...
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
+
+
+
+ Viking.prototype.receiveDamage = function(damage) {
+     var currentHealth; 
+     currentHealth = Viking.health - damage;
```
tawebbcn: `this.health`

PR #307, opened 2018-03-23
BCN Lennie Mihai Matias

Matuki92:
@cosmihai
@lenniez

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,105 @@
 // Soldier
-function Soldier() {}
+function Soldier(healthArg, strengthArg) {
+
+  this.health = healthArg;
+  this.strength = strengthArg;
+
+}
+
+Soldier.prototype.attack = function(){
+
+  return this.strength;
+
+}
+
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+
+  this.name = name;
+  this.health = health;
+  this.strength = strength;
+ // Soldier.call(this.healthArg, this.strengthArg);    /// ASK
```
tawebbcn: You should use
```
```
At the beginning of the constructor

starter-code/src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+Viking.prototype.battleCry = function (){
+
+  return "Odin Owns You All!";
+
+}
 
 // Saxon
-function Saxon() {}
+function Saxon(health, strength) {
+
```
tawebbcn: Same here. You should use Soldier.call( ... )

PR #308, opened 2018-03-23
BCN - Jorge and Pere

Interna1ta:
we didnt finished the bonus

@jorgeterol

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,79 @@
+'use strict'
+
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+}
+
+Soldier.prototype.attack = function() {
+    return this.strength;
+}
+
+Soldier.prototype.receiveDamage = function(damage) {
+    this.health -= damage;
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    Soldier.call(this, health, strength);
+    this.name = name;
+}
+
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,79 @@
...
+    this.vikingArmy.push(viking);
+}
+
+War.prototype.addSaxon = function(saxon) {
+    this.saxonArmy.push(saxon);
+}
+
+War.prototype.vikingAttack = function() {
+    var indexSaxonChoosed = Math.floor(Math.random() * this.vikingArmy.length)
+    this.saxonArmy[indexSaxonChoosed].health -= this.vikingArmy[0].strength;
```
tawebbcn: You should use the `recieveDamage` method instead

starter-code/src/viking.js
```diff
@@ -1,11 +1,79 @@
...
+}
+
+War.prototype.saxonAttack = function() {
+    var indexVikingChoosed = Math.floor(Math.random() * this.vikingArmy.length)
+    this.vikingArmy[indexVikingChoosed].health -= this.saxonArmy[0].strength;
+    if (this.vikingArmy[indexVikingChoosed].health <= 0) {
+        return this.vikingArmy.splice[indexVikingChoosed];
```
tawebbcn: how `splice` works is, one method that receive two arguments. First is the index that you want to remove, and the second one is how many element you want to remove. In this case should be like:
```
this.vikingArmy.splice(indexVikingChoose, 1);
```
Also, you don't have to return this.

PR #326, opened 2018-03-26
PAR Sami & Maggie

Sami-Az:
a little confusion when making "Viking.health" inherit from Soldier

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,66 @@
+
+
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+
+  this.health = health;
+  this.strength = strength;
+  Soldier.prototype.attack = function (){
+    return strength;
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Okay so here is a piece of code that should do the trick : 

```js 

Viking.prototype.receiveDamage = function(damage) {
  console.log(damage);
  this.health -= damage;

  if (this.health >= 0) {
    console.log(this.name + " has received " + damage + " points of damage");
  } else {
    console.log(this.name + " has died in act of combat");
  }
};

```
Using the "this" keyword allow you to refer the constructor property, it's easier to read and it resolve the error on Jasmine.

If you want to go further in the exercice, you can check what some of you did yesterday : 
https://gist.github.com/ta-web-paris/51ef24345cbf90e4ca3ecd938d032ac9

Keep up the good work :)

PR #327, opened 2018-03-26
PAR Léo W  + Jeremie

JeremieHercberg:


ta-web-paris:
Okay so first thing I have to say is that you forgot to uncomment the test, for this exercise and others it's really important for you to use the Jasmine errors, as it will help you to write clean code (and also cause it's part of the instructions ;) )

Anyway both of you were here yesterday during the extra session with Vivian, so I know that you worked hard on it! :)

Is it clear for you by the way? Or is there some points you want to work on ?

Keep up the good work !

PR #338, opened 2018-04-08
MIA-FTWD-Robert-Gelbman-lab-javascript-vikings

RobGelbman:


MarcosPalaciosIII:
good work on the vikings lab. keep up the great work. If you wan't more of a challenge, next time try actually running the code and see if you can get it to function . ;)

PR #345, opened 2018-04-09
MIA-FTWD-DiegoLeal-Vikings

DiegoLealB:


starter-code/src/viking.js
```diff
@@ -1,11 +1,71 @@
...
+}
+
+War.prototype.addViking = function(Viking){
+  this.vikingArmy.push(Viking);
+}
+War.prototype.addSaxon = function(Saxon){
+  this.saxonArmy.push(Saxon);
+}
+War.prototype.vikingAttack = function(){
+  Saxon.prototype.receiveDamage() = Viking.prototype.strength;
```
sandrabosk: Saxon's received damage will be equal to a result of Viking.attack() function. But before we get to this, we have to "pick" random Viking on which we'll call .attack() method.

sandrabosk:
Great work! Thank you for submitting.

PR #350, opened 2018-04-10
FTWD Miami - Ryan Martin - lab-javascript-vikings

ryancraigmartin:


starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+
+// // Saxon
+function Saxon(health, strength){
+  Soldier.call(this, health, strength);
+
+  Saxon.prototype = Object.create(Soldier.prototype);
+
+  Saxon.prototype.receiveDamage = function(damage){
+
+    if(health > 0){
```
MarcosPalaciosIII: `if(health > 0){` should have been `this.health` . and you should deal the damage before confirming if health > 0 not after.  `this.health -= damage` should be above the if statement to confirm if after the damage has been dealt whether the Saxon is dead or not.

MarcosPalaciosIII:
Really good work, just keep in an eye out for misspellings and positioning of functions, methods, classes, name etc. as with just 1 misplaced anything or 1 thing that is off can throw your code off its axis.   

Keep up the good work.

PR #356, opened 2018-04-11
Ftwd-miami- Brian Jaimes

BrianJaimes:
In-progress


starter-code/src/viking.js
```diff
@@ -1,11 +1,39 @@
...
+function Viking(name,healthArg, strengthArg) {
+  Soldier.call(this, healthArg, strengthArg);
+  this.name = name;
+  console.log(this.name);
+  }
+  Viking.prototype = Object.create(Soldier.prototype);
+  Viking.prototype.constructor = Viking;
+  //var viking = new Viking (name, healthArg, strengthArg);
+  
+  var viking = new Viking (name, health, strength);
```
MarcosPalaciosIII: do not need to create this. The vikings and Saxons are created in the Jasmine tests.

starter-code/src/viking.js
```diff
@@ -1,11 +1,39 @@
...
+  console.log(viking);
+  
+  
+  Viking.prototype.receiveDamage = function (damage) {
+    viking.health -= damage;
```
MarcosPalaciosIII: instead of viking.health should be this.health since your already calling the Viking.prototype

MarcosPalaciosIII:
Remember not to delete the Jasmine files or comment out the tests. the jasmine test will actually guide you through the project as it tells you what it is looking for. Jasmine also sets the variables that you will be using for the tests so when using Jasmine you will not need to create a var for viking unless its something needed for the function.  If you need further explanation just come and ask me.  You were heading in the right track and I'm sure with more time you would have finished. Keep on truckin'

B-riJ:
I get you, the Jasmine test parameters are beginning to look more familiar
now versus the binary code I was seeing it as before. Not having to create
variables unless they need to be for a function makes more sense now.
Thanks for the direction and feedback Marcos! Always appreciated. I'll keep
at it

On Fri, Apr 13, 2018 at 2:42 PM, MarcosPalaciosIII <notifications@github.com

PR #412, opened 2018-06-04
PAR Jihad

eloji9:
I pretty much used the lesson of this morning to get the exercise done.
I'm sure that with more practice, I'll be able to remember the commandes.

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,104 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+}
+
+Soldier.prototype.attack=function(){
+    return this.strength;
+}
+
+Soldier.prototype.receiveDamage=function(damage){
+    this.health -= damage;
+}
 
 // Viking
-function Viking() {}
+Viking.prototype=Object.create(Soldier.prototype);
+
+function Viking (name,health,strength){
+    this.name=name;
+    this.health=health;
+    this.strength=strength;
+}
+
+Viking.prototype.attack=function(){
+    return this.strength;
+}
+
+Viking.prototype.receiveDamage=function(damage){
+    this.health -= damage;
+    if (this.health > 0){
+        return this.name + " has received " + damage + " points of damage";
+    }
+    if (this.health <= 0){
+        return this.name + " has died in act of combat";
+    }
+}
+
+Viking.prototype.battleCry = function(){
+    return "Odin Owns You All!"
+}
 
 // Saxon
-function Saxon() {}
+Saxon.prototype=Object.create(Soldier.prototype);
+
+function Saxon (health,strength){
+    this.health=health;
+    this.strength=strength;
+}
+
+Saxon.prototype.attack=function(){
+    return this.strength;
+}
+
+Saxon.prototype.receiveDamage=function(damage){
+    this.health -= damage;
+    if (this.health> 0){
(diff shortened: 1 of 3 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,104 @@
...
 // Viking
-function Viking() {}
+Viking.prototype=Object.create(Soldier.prototype);
+
+function Viking (name,health,strength){
+    this.name=name;
+    this.health=health;
+    this.strength=strength;
+}
+
```
ta-web-paris: Here you basically recreated the Viking from scratch, while the idea was to import characteristics from the Soldier constructor.
In this case it's not a big deal, because the Soldier is pretty simple, but sometimes you'll face super complex constructor functions that you don't want to be copying line by line ;)

The idea was to use the .call method, as follows:

```javascript
function Viking( name, health, strength ) {
    Soldier.call( this, health, strength );
    this.name = name;
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,104 @@
...
 
 // Saxon
-function Saxon() {}
+Saxon.prototype=Object.create(Soldier.prototype);
+
+function Saxon (health,strength){
+    this.health=health;
+    this.strength=strength;
+}
+
```
ta-web-paris: Same remark than for the Viking, the Saxon constructor should have looked like this:
```javascript
function Saxon( health, strength ) {
    Soldier.call( this, health, strength );
};
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,104 @@
...
+War.prototype.vikingAttack = function(){
+    var attack = this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)].receiveDamage(this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)].strength);
+    if (this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)].health <= 0){
+        this.saxonArmy.splice(0,1);
+    }
+    return attack;
+}
+
```
ta-web-paris: Here, you vikingAttack and saxonAttack methods work, but there's two things:
1. you repeat the same long lines a couple times, which makes room for type errors
2. updating / modifying such code takes longer, because you need to makes the changes in several places instead of one
3. it's harder to read for someone that doesn't know your code, and making your functions easy to understand is a major key in teamworking with developers

A way to solve those problems, would be to capture all the elements you use inside properly-named variables so you know exactly what you're doing, and it's way clearer. Here's an example:
```javascript
War.prototype.vikingAttack = function() {
    
    var randomSaxNum = Math.floor( Math.random() * this.saxonArmy.length );
    var randomSaxon = this.saxonArmy[ randomSaxNum ];
    
    var randomVikNum = Math.floor( Math.random() * this.vikingArmy.length );
    var randomViking = this.vikingArmy[ randomVikNum ];

    var result = randomSaxon.receiveDamage( randomViking.attack() );

    if( randomSaxon.health <= 0 ) {
        this.saxonArmy.splice( randomSaxNum, 1 );
    }

    return result;
};
```

Last thing, in case the Viking / Saxon dies, you executed the splice on the index 0 every time. Which means if Viking / Saxon dies, you always splice the first in the array. The idea here is to splice the viking that just died, so you want to replace the 0 by the actual index in the armies array, as in the example I pasted above

ta-web-paris:
I added a couple remarks, but you generally well understood how it works :)
Keep going!

PR #413, opened 2018-06-04
PAR Shanshan

moonchie:
Bonus war part didn't finish.

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,97 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+};
+
+Soldier.prototype.attack = function (){
+    return this.strength
+}
+
+Soldier.prototype.receiveDamage = function (dama){
+    this.health = this.health - dama;
+}
+
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    Soldier.call(this, health, strength);
+    this.name = name;
+};
+
+var newViking = new Viking("Brynhild", 100, 300);
+console.log(newViking);
+console.log(newViking.strength);
+
+Viking.prototype = Object.create(Soldier.prototype);
+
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+
+//test field
+function War() {
+
+// Empty army
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+
+// Add a viking and a Saxon to the arty
+    war.prototype.addViking = function(viking){
```
ta-web-paris: Here you create a prototype for "war", but the constructor you created was called "War".
Javascript is case-sensitive, so be wary of how you spelled it the first time ;)
(Also, Constructor Functions Should Always Have A Big First Letter)

starter-code/src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+        this.vikingArmy.push(viking);
+    };
+
+    war.prototype.addSaxon = function(saxon){
+        this.saxonArmy.push(saxon);
+    };
+
+    //Random select one
+    var randomViking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
```
ta-web-paris: This is good and it actually works, but you might want to do it in two steps, so it's clearer to read and easier to update if you want to change something. I could suggest something like this:
```javascript
War.prototype.saxonAttack = function() {
    
    var randomSaxNum = Math.floor( Math.random() * this.saxonArmy.length);
    var randomSaxon = this.saxonArmy[randomSaxNum];
    
    var randomVikNum = Math.floor( Math.random() * this.vikingArmy.length);
    var randomViking = this.vikingArmy[randomVikNum];

    var result = randomViking.receiveDamage( randomSaxon.attack() );

    if( randomViking.health <= 0 ) {
        this.vikingArmy.splice( randomVikNum, 1 );
    }

    return result;
};
```

ta-web-paris:
The bonus question with war is quite tricky, but you understood the lesson well, congrats!

- Lucas

PR #422, opened 2018-06-06
PAR - Paul de Barbeyrac

Paul-de-Barbeyrac:


starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+
+War.prototype.vikingAttack = function () {
+  
+  function entierAleatoire(min, max) {
+    return Math.floor(Math.random() * (max - min + 1)) + min;
+  }
+
+  var RandomViking = this.vikingArmy[entierAleatoire(0, this.vikingArmy.length - 1)];
+  var RandomSaxon = this.saxonArmy[entierAleatoire(0, this.saxonArmy.length - 1)];
+
```
M-Gadd: instead of creating a function that you have to give 2 arguments and recall it to bring back a random number, you can just generate a random number based on the length of the array, save it in a variable and use the number every time as an index on the array you need to use.

starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+}
+
+War.prototype.addViking = function (Viking) {
+  this.vikingArmy.push(Viking);
+};
+
+War.prototype.addSaxon = function (Saxon) {
+  this.saxonArmy.push(Saxon);
+};
+
```
M-Gadd: your push functions are fine

starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+// Status of War
+War.prototype.showStatus = function() {
+    if (this.vikingArmy.length > 0 && this.saxonArmy.length > 0) {
+      return "Vikings and Saxons are still in the thick of battle.";
+    } else if (this.saxonArmy.length === 0) {
+      return "Vikings have won the war of the century!";
+    } else if (this.vikingArmy.length === 0) {
+      return "Saxons have fought for their lives and survive another day...";
+    } 
+  }
```
M-Gadd: good job

PR #423, opened 2018-06-06
PAR Hanane MOUALIL

HananeDUP:


starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
 // Viking
-function Viking() {}
+function Viking(name,health,strength) {
+    Soldier.call(this,health,strength);
+    this.name=name;
+    
+}
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
```
M-Gadd: You don't need this here. you already inherited the Solider class

starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+Viking.prototype.battleCry=function(){
+    return "Odin Owns You All!";
+}
 
+Viking.prototype.receiveDamage=function(damage){
+    this.health-=damage;
+    if(this.health-damage>0){
```
M-Gadd: This probably will not work because after you deduct the damage from the health you are checking if the new health - damage will be > 0 which means that you deducting double the damage.

starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+    var health   = 300;
+
+    return new Viking(name, health, strength);
+  }
+
+  function generateSaxon () {
+    var health   = 60;
+    var strength = 25;
+    return new Saxon(health, strength);
+  }
```
M-Gadd: from line 55 to line 67 are not needed, because you already have your variables before and you don't need to generate anything using manual entries. When you put manual entries on the property it will always return those values regardless of the arguments you give the function

starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+      var v=Math.floor(Math.random()*(war.vikingArmy.length));
+      var s=Math.floor(Math.random()*(war.saxonArmy.length));
+  
+      var myviking=war.vikingArmy[v];
+      var poorsaxon=war.saxonArmy[s];
+      var dmg=myviking.strength;
+      // poorsaxon.receiveDamage(myviking.attack());
+     return poorsaxon.health-dmg;
+  
+  }
```
M-Gadd: Ok this whole function will not work because you are trying to generate new vikings and saxons inside your function.... what you need to do after you already pushed to the arrays of vikingArmy and saxonArmy  is to generate 2 random numbers that you use as an index on both arrays and they you call the receive damage function of a random one on the other random one depends on who is attacking whom. 

so if a viking is attacking a Saxon then you need call the receive damage function on the RandomSaxon giving it the strength of the RandomViking .. and if the RandomSaxon's health is less than the RandomViking's strength, then you have to remove him from the array as he will be dead. Same thing the other way around.

starter-code/src/viking.js
HananeDUP: I remove it

starter-code/src/viking.js
HananeDUP: You're right! I did it right for saxon but not viking :)

starter-code/src/viking.js
HananeDUP: I agree. I put it because i was working in repl to test the values...Now I remove them

starter-code/src/viking.js
HananeDUP: I worked again the methods. Now it will work (I hope :))

PR #426, opened 2018-06-07
PAR - Robert Vivanco

rovisal:
Ongoing


starter-code/src/viking.js
```diff
@@ -1,11 +1,48 @@
...
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+  this.name = name;
+  Soldier.call(this, health, strength);
+}
+Viking.prototype = Object.create(Soldier.prototype);
+
+Viking.prototype.attack = function() {
+  return this.strength;
```
M-Gadd: you don't need this prototype because you are already inheriting it from the soldier constructor.

starter-code/src/viking.js
```diff
@@ -1,11 +1,48 @@
...
+
 
 // Saxon
-function Saxon() {}
+function Saxon(health, strength) {
+  Soldier.call(this, health, strength);
+}
+Saxon.prototype.attack = function() {
+  this.strength;
+}
```
M-Gadd: Same as viking

PR #427, opened 2018-06-07
PAR - Pauline Ronez

paulinezenor:


starter-code/src/viking.js
```diff
@@ -1,11 +1,66 @@
...
+        return ("A Saxon has received " + damage + " points of damage")}
+        else return ("A Saxon has died in combat")
+
+    }
+}
+
 
 // War
 function War() {}
+
```
M-Gadd: here you need to create 2 arrays, one for the viking army and one for the Saxon army and then 2 function 2 push them inside the array 

inside the war function you need to have 2 random numbers and choose a random viking and a random Saxon then have one of them receive damage from the strength of the other on, then check if he dies or not.

PR #453, opened 2018-06-18
BCN Smashed it, Will and Sebastian

se66e:
@dambusm

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,99 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+}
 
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+};
+
+War.prototype.addSaxon = function(saxon) {
+  this.saxonArmy.push(saxon);
+};
+
+War.prototype.vikingAttack = function() {
+  var randomNumber = Math.floor(Math.random() * this.saxonArmy.length);
+  var saxon = this.saxonArmy[randomNumber];
+  var randomNumber2 = Math.floor(Math.random() * this.vikingArmy.length);
```
tawebbcn: There's some repeated code here. Could we do a helper function to help with this?

PR #456, opened 2018-06-18
BCN - Kahlil & Julio Vikings

Lilhak4:
@juliocarlosvelez


starter-code/src/viking.js
```diff
@@ -1,11 +1,80 @@
 // Soldier
-function Soldier() {}
+function Soldier(h, s) {
+  this.health = h;
+  this.strength = s;
```
tawebbcn: Try to use the full name for health and strength. Readability is key!

PR #460, opened 2018-06-18
MIA- Ale Aragao

arag021:


MarcosPalaciosIII:
saw your code on your partners pull req. For some reason you changed something that does not allow specRunner to check the work your doing. Remember when doing test driven coding to not change any of the test files or folders. You should only modify the files and folders given to you to change.  
Overall excellent job on the vikings lab. 

keep up the excellent work.

PR #465, opened 2018-06-19
MIA FT 061118 Alina Kitaeva

akitaeva:
one failure left =/


starter-code/src/viking.js
```diff
@@ -1,11 +1,160 @@
...
+    var saxonIndex = Math.floor(Math.random()*this.saxonArmy.length);
+    var someViking = this.vikingArmy[vikingIndex];
+    var someSaxon = this.saxonArmy[saxonIndex];
+    
+
+    var result = someSaxon.receiveDamage(someViking.attack());
+    if (someSaxon.health <=0) {
+        this.saxonArmy.splice(saxonIndex,1);
+    }
+    else {
```
MarcosPalaciosIII: you don't need the else here since you need to always return the result after an attack

starter-code/src/viking.js
```diff
@@ -1,11 +1,160 @@
...
+    var vikingIndex = Math.floor(Math.random()*this.vikingArmy.length);
+    var saxonIndex = Math.floor(Math.random()*this.saxonArmy.length);
+    var someViking = this.vikingArmy[vikingIndex];
+    var someSaxon = this.saxonArmy[saxonIndex];
+
+    var result = someViking.receiveDamage(someSaxon.attack());
+    if (someViking.health <=0) {
+        this.vikingArmy.splice(vikingIndex,1);
+    }
+    else {
```
MarcosPalaciosIII: you don't need the else here since you need to always return the result after an attack

MarcosPalaciosIII:
excellent work on the vikings lab. Just remember that you don't always need to do an else for an if statement.

PR #481, opened 2018-08-04
Ber-Nhan

ntnhan30:


ta-web-ber:
`War` class missing, but good so far.

Silvio

PR #486, opened 2018-08-05
BER-Dalina-done

DalinaLeila:


ta-web-ber:
`War` class is missing, but the previous iterations are very good.
Nice job.

Silvio

PR #509, opened 2018-08-13
PAR Julie & Pauline

paulettepaillette:
@youlie03

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,68 @@
 // Soldier
-function Soldier() {}
+function Soldier(healthArg, strengthArg) {
+
+        this.health = healthArg;
+        this.strength = strengthArg;
+        this.attack = function() {
+            return(this.strength);
+        };
+        this.receiveDamage = function(theDamage) {
+            this.health -= theDamage;
+        }
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    this.name = name;
+    Soldier.call(this, health, strength);
+    this.receiveDamage = function(theDamage) {
+
+        this.health -= theDamage;
+
+        if (this.health > 0){
+            return (this.name + " has received " + theDamage + " points of damage");
+        }
+        else {
+            return (this.name + " has died in act of combat");
+        }
+    }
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,68 @@
 // Soldier
-function Soldier() {}
+function Soldier(healthArg, strengthArg) {
+
+        this.health = healthArg;
+        this.strength = strengthArg;
```
ta-web-paris: Here, you properly define the health and strength characteristics, but this is not how you should define the methods. The way you did it works, but there's a more "organized" way of doing it:

```javascript
function Soldier(healthArg, strengthArg) {
    this.health = healthArg;
    this.strength = strengthArg;
}

    return this.strength;
}

Soldier.prototype.receiveDamage = function( damageArg ) {
    this.health -= damageArg;
}
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,68 @@
...
+            this.health -= theDamage;
+        }
+}
 
 // Viking
-function Viking() {}
+function Viking(name, health, strength) {
+    this.name = name;
+    Soldier.call(this, health, strength);
+    this.receiveDamage = function(theDamage) {
```
ta-web-paris: And then here, your Viking function should look like this:

```javascript
function Viking(name, health, strength) {
    this.name = name;
}

// You create the viking's prototype based on the Soldier's one
Viking.prototype = Object.create(Soldier.prototype);

// You redefine the methods that you want to modify
Viking.prototype.receiveDamage = function(theDamage) {

    this.health -= theDamage;

    if (this.health > 0){
        return (this.name + " has received " + theDamage + " points of damage");
    }
    else {
        return (this.name + " has died in act of combat");
    }
}

    return ("Odin Owns You All!");
}
```

ta-web-paris:
The exercise works as you wrote it, but organizing the method declaration like we did is pretty important too. But you're almost there! ;)

The beginning of the War function was ok, but the rest is a bit more tricky hehe. Anyway you'll have time to train this type of thinking later in the bootcamp so don't worry :)

Good job :)
Lucas

PR #510, opened 2018-08-13
PAR Aurélien and Abi 

abisummers:
@LOOFFI

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,75 @@
 // Soldier
-function Soldier() {}
+function Soldier(healthArg, strengthArg) {
+  this.health = healthArg;
+  this.strength = strengthArg;
+}
+
+Soldier.prototype.attack = function() {
+  return this.strength;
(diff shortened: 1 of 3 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,75 @@
...
+};
 
 // Viking
-function Viking() {}
 
+//constructor function
+function Viking(name, health, strength) {
+  this.name = name;
+  this.health = health;
+  this.strength = strength;
```
ta-web-paris: When making the Viking inherit from the Soldier, the idea is to not write his strength and health characteristics again, but to take them from the Soldier. Here there's only two items, but there can be way more, and they can also be more complicated.
So there's the way to procede:

```javascript
function Viking(name, health, strength) {
  this.name = name;
}
```

And same for the Saxon :)

ta-web-paris:
Function and inheritance is generally well understood, good :)

Good job :)
Lucas

PR #512, opened 2018-08-13
PAR John & Paul

PaulAbarnou37:
@johndimitroff


ta-web-paris:
Clean code and all, nothing to say.

The War function is a bit more tricky though, but you'll have time to train this type of thinking later in the bootcamp so don't worry :)

Good job :)
Lucas

PR #589, opened 2018-10-02
[PAR-PT] Justine & Merwan

Mrwn211:
jusque fonction receiveDamage sur viking
@ta-web-paris

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,8 +1,36 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+  this.attack = function() {
+    return this.strength;
+  };
+  this.receiveDamage = function(damage) {
+    this.health = this.health - damage;
+  };
+}
 
 // Viking
-function Viking() {}
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,8 +1,33 @@
...
+Viking.prototype.constructor = Viking;
+
+function Viking(name, health, strength) {
+  Soldier.call(this, health, strength);
+  this.name = name;
+};
+Viking.prototype.receiveDamage=function (damage){
+  Soldier.__proto__.receiveDamage(damage);
+  if (this.health > 0) {
+      return name + " has received " + damage + " points of damage";
```
ta-webpt-par: The `name` here belongs to the Viking instance, you need to use `this.name`

starter-code/src/viking.js
```diff
@@ -1,8 +1,33 @@
...
+
+Viking.prototype = Object.create(Soldier.prototype);
+Viking.prototype.constructor = Viking;
+
+function Viking(name, health, strength) {
+  Soldier.call(this, health, strength);
+  this.name = name;
+};
+Viking.prototype.receiveDamage=function (damage){
+  Soldier.__proto__.receiveDamage(damage);
```
ta-webpt-par: You could do this like this `Soldier.prototype.receiveDamage.call(this, damage);`

PR #592, opened 2018-10-02
<PAR PT> Laetitia et Sandra

SanDDupt:
Soldier + Viking
@ta-webpt-par

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,8 +1,34 @@
 // Soldier
-function Soldier() {}
+function Soldier(health, strength) {
+    this.health = health;
+    this.strength = strength;
+    this.attack = function() {
+        return this.strength;
+    };
+    this.receiveDamage = function(damage) {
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,8 +1,34 @@
...
+
+    this.receiveDamage = function(damage) {
+        this.health = health - damage;
+        if (this.health > 0) {
+            return name + " has received " + damage + " points of damage";
+        }
+        else {
+            return name + " has died in act of combat";
+        }
+    };
```
ta-webpt-par: you are missing the function battleCry() to be done with the Viking

starter-code/src/viking.js
```diff
@@ -1,8 +1,34 @@
...
+}
 
 // Saxon
 function Saxon() {}
```
ta-webpt-par: Use the same logic for the Saxon.

PR #655, opened 2018-10-27
[MAD DE] Fernando Pérez de Ayala

ferpar:
Just so you know I completed this, whether you check it is up to you of course. All Iterations and the Bonus have been completed. Forgot to commit on every iteration though ^^'... and thanks a lot for all your previous feedback! ^^

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,119 @@
 // Soldier
-function Soldier() {}
+function Soldier(healthArg, strengthArg) {
+    this.health = healthArg
+    this.strength = strengthArg
+}
+
+Soldier.prototype.attack = function () {
+    return this.strength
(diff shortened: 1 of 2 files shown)


ferpar:
Ooops.. now I realized this is the first time I pull request for this one. I got confused with the DE of Thursday. For that one I do intend to make a second pull request.

ta-web-mad:
Awesome exercise. But be careful:
``` 
this.vikingArmy = this.vikingArmy.filter(function (viking) {
        return (viking.health > 0)
    });
```
Every time there's an attack, you have to look all the items inside the Array. Think if you have 1 million elements inside the array. There's other better ways to do it. See the ```splice method``` with the index. 

Even so, well done.

#checked #Teo :muscle:

PR #656, opened 2018-10-28
MIA FTWD Rachel Hofer - Vikings JS

Rachel-Hofer:


MarcosPalaciosIII:
Good work on the vikings lab. You can change up your function which splices out the deceased viking or saxon once they drop to 0 hp or below if you deal damage and then run the check for the hp. If by chance the function is running too fast, you can usually put the loop to check for the hp in a setTimeout function. 

Keep up the good work.

PR #657, opened 2018-10-28
MIA FTWB Rustam Samigullin ver 1.1

Rustam1993:


MarcosPalaciosIII:
Good work on the vikings lab. On the attack function that splices the dead viking or saxon, you should be checking last to see which viking or saxon you will splice from the array. If that part runs to fast then you can put a setTimeout function in order to have it run when you want it too. 

Keep up the good work.

MarcosPalaciosIII:
Great work on the vikings lab. You have everything working they way it should and the dead viking or saxon is removed from the array the moment their dead. 

Keep up the great work.

PR #676, opened 2018-10-29
MIA FTWD Isaiah Gonzalez

Isaiah361:


MarcosPalaciosIII:
Good work on the vikings lab. Remember when coding to check for spelling when creating functions and classes. Misspelled variables and or words can usually make your code not work, especially if it's a crucial variable to the function or class. Also, it's good practice to test your code prior to setting to deployment to help you catch those bugs as well as to assure that you are not missing anything. 

Keep up the good work.

PR #739, opened 2018-11-22
[MIAPTOCT11] Chris Rodriguez

Jackslaps:
…ar (only added push functions for viking and saxons)


ta-web-mia:
You finished almost all of it and it looks good! still missing the last parts of the war constructor but this is good enough to pass :)
- Francisco

PR #808, opened 2019-01-15
Paris - Christophe & Céline

frietpoune:
Finished :) 

Let war begin...

@grosbegue


ta-web-paris:
What you've done is good! You seem to understand what we covered in class. I like how you've used a lot of console logs - they're going to be super important especially during project weeks. 👏

PR #810, opened 2019-01-15
Paris Aurelie et Samuel

AurelieBeno:
@petitkriket


ta-web-paris:
What you've done is good! I really like how you've added comments to your code to explain what you're doing. You seem to understand what we've covered in class 🙌

PR #811, opened 2019-01-15
Paris: Poeli & Raja

RaproCode:
We couldn't finish all because we were stuck a few times with The correct solution, but typos here and there so Jasmine didn't give the green light. Overall, we understood quite well

@pbojork


starter-code/src/viking.js
```diff
@@ -1,8 +1,47 @@
...
+    this.health = this.health - damage 
+  } 
+}
+
 
 // Viking
-function Viking() {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(name, health, strength);
```
ta-web-paris: You could have simplified this a little 

``` 
    super(health, strength);
    this.name = name;
  } 
```

The health and strength is getting passed from the Solider function

ta-web-paris:
What you've done is good, and you've passed all the tests for the iterations you completed! If you'd like to continue the lab, let us know and we can help you or give you more feedback :)

PR #812, opened 2019-01-15
Paris: Francesca, Anoop

brwnrcluse:
done except for bonus

@wowfranzik

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,80 @@
 // Soldier
-function Soldier() {}
+class Soldier {
+  // make constructor
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,80 @@
...
+    };
+  }
+}
 
 // Viking
-function Viking() {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = "Harald";
```
ta-web-paris: Here it should be `this.name = name` You didn't need to hardcode the value

ta-web-paris:
What you've done is good! You seem to have understood the lesson! 💪


If you want to finish the bonus (you don't have to) then let us know so we can give you some feedback.

PR #814, opened 2019-01-15
Paris Nadje & Delphine

Deie:
@NadjieVal

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,92 @@
 // Soldier
-function Soldier() {}
+class Soldier {
+    constructor(health, strength) {
+        //console.log("Your soldier health is " + health + " and his strength is " + strength);
+        this.health = health;
+        this.strength = strength;
+       // this.damage = health --;
+    }
+
+    attack() {
+        return this.strength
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,92 @@
...
+  }
+  addViking(Vicking) {
+    this.vikingArmy.push(Vicking);
+  }
+
+  addSaxon(Saxon) {
+    this.saxonArmy.push(Saxon);
+  }
+    vikingAttack() {
+        Saxon.receiveDamage() = vicking.strength();
```
ta-web-paris: For this function to work and pass all the tests you need to:

-get a random Viking from your viking array 
-get a random Saxon from your saxon array 
-When it says "should make a Saxon receiveDamage() equal to the strength of a Viking" this just means that when the viking is attacked the damage points is equal to the strength of the saxon. You need to include something like this  `randomViking.receiveDamage(randomSaxon.strength);`
- you need to slice the array when the saxon dies

ta-web-paris:
What you've done is good! If you want to finish the exercise then let us know so we can give you more feedback on it 🙌

PR #834, opened 2019-01-29
Henry Doce - FTWD - Miami

hendo12:


ta-web-mia:
Good Job Henry!!!
in order to pass the last two test you would need to do the following:
```
vikingAttack() {
    const randomSaxon = this.saxonArmy[
      Math.floor(Math.random() * this.saxonArmy.length)
    ];
    const randomViking = this.vikingArmy[
      Math.floor(Math.random() * this.vikingArmy.length)
    ];
    const saxonDamage = randomSaxon.receiveDamage(randomViking.strength);

    // if (randomSaxon.health === 0) { ----- since the health can also go down to a negative number
    // you need to make the condition less than or equal in order to always work
      this.saxonArmy.splice(randomSaxon);
    } else {
      return saxonDamage;
    }
  }
```

PR #860, opened 2019-02-07
[AMS] Merle Jansen

jurlem:
Vikings without War (bonus)

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,144 @@
 // Soldier
-function Soldier() {}
 
+function Soldier(health, strength) {
+  this.health = health;
+  this.strength = strength;
+};
+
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,144 @@
...
+    this.saxonArmy.splice(randSaxonIndex, 1); // Find the index of "randSaxon", then remove element from that position                                                            
+  }
+  return vikingAttackingSaxon;
+}
+
+/////
+
+
+War.prototype.saxonAttack = function () {
+  var randVikingIndex = Math.random(()*this.vikingArmy.length);
```
TA-Ironhack-Matt: The '()*' before this.vikingArmy is causing an error in your console.  

If you remove it your code will run correctly :)

PR #925, opened 2019-03-25
PAR alex

alexandfox:
completed Vikings lab + war

please check, thank you!


starter-code/src/viking.js
```diff
@@ -1,11 +1,126 @@
...
+		this.strength = strength;
+	};
+
+	attack() {
+		return this.strength;
+	}
+
+	receiveDamage(damage) {
+		this.health -= damage;
+		return;
```
ta-web-paris: You don't need to return here.

ta-web-paris:
Not much to say, you really got it ! Bravo

PR #927, opened 2019-03-26
PARIS Daniel Diaz

DaniTOLETUM:


ta-web-paris:
Not much to say. It is good. I like the fact that you push on your repo after each iteration. It is a really good practice.

PR #932, opened 2019-03-26
PAR - Susan

sioucre:


starter-code/src/viking.js
```diff
@@ -1,11 +1,119 @@
...
+    this.strength = strength;
+  }
+
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+    return;
```
ta-web-paris: No need to return here.

ta-web-paris:
Clean code, everything works ! Bravo

PR #954, opened 2019-04-09
MIA FTWD - Angel Biondi

AngelBiondi:
done!


ta-web-mia:
Good job! Hopefully this gives you a better handle on classes and how it relates to object oriented programming. Keep up the good work!

PR #958, opened 2019-04-09
MIA FTWD APR-Katy Brown

katybrown91:


ta-web-mia:
Nice job! In the War function, you need to use the array push() method to add vikings or saxons to their respective army arrays. You would use += when dealing with some sort of number, but arrays are another thing entirely. Hopefully this helped you get a better grasp on how to use classes, and how they relate to object oriented programming in general. Keep it up!

PR #962, opened 2019-04-09
MIA FTWD - Nour Cherif Essoussi - April 2019

Nourcs:
viking.js ( ES6 Classes )
vikingVanilla.js ( Functions )


ta-web-mia:

PR #1037, opened 2019-06-14
MIA-FTWD-Sergio

sergiovelo17:


MarcosPalaciosIII:
Excellent job on the viking lab. Object oriented programming is something that is used in order to keep your code clean and allows you to keep functions more compartmentalized. This is also good when setting up your schema in your back end development, since you can have certain types of user (for example) to be allowed certain functionality in an app.  Most gaming companies use this for the different classes that they have to choose from in order to keep abilities better separated to each combat style. With more practice you will find that this gets easier and makes your code cleaner and easier to understand. 

Keep up the excellent work.

PR #1043, opened 2019-06-14
MIA-FTWD-Vikings-Pablo Vazquez

PabloVazquezFayat:


starter-code/src/viking.js
```diff
@@ -1,11 +1,121 @@
...
+}
+
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength){
+        super(health, strength);
+        this.name = name;
+        this.health = health;
```
ta-web-mia: When you use super, you don't have to include this.x = x; the super automatically does that for you

ta-web-mia:
Other than the super, everything is laid out nice. Good job!

PR #1140, opened 2019-08-14
PAR FTWD DORO - done

DoriDoro:


ta-web-paris:
Great! You just need to replace "A Saxon has died in act of combat" with "A Saxon has died in combat" for Jasmine to be all green :)

PR #1211, opened 2019-08-28
Vanessas VikingLab

VanessaAhrens:


marianamv112:
Hi Vanessa.
Please be careful when pushing your projects. Your code has errors and that makes it not pass any test. Also I would recommend you to check out the correct syntax for the return statement.
Besides that, your lab is just fine. The TAs are always available if you need some help to finish the Bonus iteration.

VanessaAhrens:
Hi Mariana,

I will try it, thx for your Feedback. May you explain me tomorrow, what you mean with “push careful”? I have still from time to time Problems to find the correct repo for pull request. Or do you really mean the push in Terminal?

See you tomorrow ☺

Von: marianamv112 [mailto:notifications@github.com]
Gesendet: Mittwoch, 28. August 2019 16:05
An: ironhack-labs/lab-javascript-vikings
Cc: Ahrens, Vanessa; Author
Betreff: Re: [ironhack-labs/lab-javascript-vikings] Vanessas VikingLab (#1211)


Hi Vanessa.
Please be careful when pushing your projects. Your code has errors and that makes it not pass any test. Also I would recommend you to check out the correct syntax for the return statement.
Besides that, your lab is just fine. The TAs are always available if you need some help to finish the Bonus iteration.

marianamv112:
Yes, I mean to push it in terminal. When you do a push, you're sending your code to the github repo in the status it is. So if you're code has some error, or is not running or something like that, when someone clones your repo we'll receive a project than is not working. This can be a problem when your working professionally. So it's always better to comment the parts of your code that ar not working and write a note about it (you can add it the commit message for example) even if that means you're pushing an unfinished work ;)

PR #1216, opened 2019-08-28
MIA-ftwd-Aug-2019 Done

marven250:


ta-web-mia:
Nice Job! This lab is a great reference further on in the cohort, so it's good to understand classes and constructors.

PR #1217, opened 2019-08-28
MIA FTWD Jorge Luis

jorgeluis42:


ta-web-mia:
Nice Job! If you can understand the fundamentals of classes in JS, you will be sitting pretty for the next module.

PR #1221, opened 2019-08-28
mia ftwd Vitor Tomazzi

VitorTomazzi:


ta-web-mia:
Nice Job! This is a great reference for OOP and classes, so I'm glad you're able to understand the fundamentals of this exercise.

PR #1222, opened 2019-08-28
MIA-FTWD-Alejandro Gomez

alegomez1:


ta-web-mia:
It looks good for what is finished, but it would really benefit you to do the War class. This entire lab will be a great reference as we go deeper into JS

PR #1224, opened 2019-08-28
MIA-FTWD-Enrique Narciso

e-narciso:
passes all Jasmine tests


ta-web-mia:
Good work! Understanding OOP will be one of  the cornerstones of JS programming

PR #1225, opened 2019-08-29
MIA FTWD FRANCESCA

Ladyfancy:


ta-web-mia:
So far it looks good, but try to take the time to finish the War class. This is a great exercise in understanding classes and OOP

PR #1226, opened 2019-08-29
MIA-FTWD Christian Roque

ChristianRoque:


ta-web-mia:
Good Job! You seem to know all the fundamentals of OOP, which is one of the best skills to have as a JS developer

PR #1280, opened 2019-10-22
[PAR FTWD] - Amine

aminelahrichi:


starter-code/src/viking.js
```diff
@@ -1,11 +1,124 @@
...
+  saxonAttack() {
+    var randomSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+    var chosenSaxon = this.saxonArmy[randomSaxon];
+    var randomViking = Math.floor(Math.random() * this.vikingArmy.length);
+    var chosenViking = this.vikingArmy[randomViking];
+
+    chosenViking.receiveDamage(chosenSaxon.strength);
+    if (chosenViking.health <= 0) {
+      this.vikingArmy.splice(randomViking, 1);
+      return `${chosenViking.name} has died in act of combat`;
```
ta-web-paris: To avoid to rewrite the sentence (which is the return of the function receiveDamage() ), you could hold this function into a variable(it will at the same time call the function and save the result) and return this variable at the end.

ta-web-paris:
Great Job!

PR #1283, opened 2019-10-22
[PAR_FTWD] Nicolas

Empty-n:
Thanks Amine, for the random thing !
Everything works fine :)

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,93 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,93 @@
...
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+  receiveDamage(damage) {
+    this.health -= damage;
```
ta-web-paris: because the function receiveDamage is already define in Soldier, you could do :
receiveDamage(damage){
   super.receiveDamage(damage)
   ....add your conditions with the sentences
}

ta-web-paris:
Good Job ;)

PR #1284, opened 2019-10-22
[PAR WDFT] Nouhad - Viking 

nouhadk:
First iterations, without bonus

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,61 @@
 // Soldier
-class Soldier {}
+class Soldier {
+constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+}
+attack() {
+    return this.strength 
+}
+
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,61 @@
...
+class Viking extends Soldier{
+    constructor(name, health, strength){
+        super(health, strength)
+        this.name = name;
+    }
+    attack() {
+        return this.strength
+    }
+receiveDamage(damage) {
+    this.health -= damage
```
ta-web-paris: because it'a an extend class of soldier and receiveDamage is already define in Solider, you could write: 
receiveDamage(damage){
  super.receiveDamage(damage)
  ... add your condition and sentences
}

ta-web-paris:
Good job but try to challenge yourself by doing the bonus because all the difficulties are in it.

PR #1287, opened 2019-10-22
[PAR WDFT] Andy + Hanna

hanna974:
done

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,92 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+
+    attack() {
+        return this.strength;
+    }
+
+    receiveDamage(damage) {
+        this.health = this.health - damage;
+    }
+}
 
 // Viking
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,92 @@
...
+    addSaxon(saxon) {
+        this.saxonArmy.push(saxon);
+    }
+
+    vikingAttack() {
+        let randomNumberSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+        var randomSaxon = this.saxonArmy[randomNumberSaxon];
+        var randomViking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
+        if (randomViking.strength >= randomSaxon.health) {
+            this.saxonArmy.splice(randomNumberSaxon, 1);
```
ta-web-paris: Normally here you are supposed to call receiveDamage(damage) before removing the dead solider form the array...
To do that you could hold your function inside a variable (that will at the same time call your function and save the result) and then return this variable at the end.

ta-web-paris:
Great Job Team!
"Odin Owns You All!"

PR #1291, opened 2019-10-22
[PAR FTWD]- LOUISE

LouiseMarissal:
done except the all bonuses

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,63 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,63 @@
...
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+  receiveDamage(theDamage) {
+    this.health -= theDamage;
```
receiveDamage(damage){
  super.receiveDamage(damage)
}

ta-web-paris:
Good job but try to do the bonus!
 It will challenge yourself and I'm sure you can do it Viking!

PR #1298, opened 2019-10-22
BER-WDFT-Fernando

feraraujofilho:


starter-code/src/viking.js
```diff
@@ -1,11 +1,119 @@
...
+  receiveDamage(theDamage) {
+    this.health -= theDamage;
+  }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super();
```
itstheandre: Both here and on the Saxon, if you used the super, you wouldnt need to declare this.health and this.strength ,becasue those properties would come from the parent class.

PR #1299, opened 2019-10-22
[BER-FTWD19-Hanna]

Lorelore22:


starter-code/src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+        this.vikingArmy.push(viking);
+    }
+    addSaxon(saxon) {
+        this.saxonArmy.push(saxon);
+    }
+    vikingAttack() {
+        let randomNr1 = Math.floor(Math.random() * this.saxonArmy.length);
+        let randomNr2 = Math.floor(Math.random() * this.vikingArmy.length);
+        let result = this.saxonArmy[randomNr1].receiveDamage(this.vikingArmy[randomNr2].strength);
+        if (this.saxonArmy[randomNr1].health < 0) {
```
itstheandre: In this case, you're getting an error, because nothing is happening if a saxon or a viking has health of 0. Only if it's health goes below 0.

You just need to add an equals and all the tests will pass :)

PR #1300, opened 2019-10-22
[BER-FTWD19-Laura]

ning3l:


starter-code/src/viking.js
```diff
@@ -1,11 +1,131 @@
...
+    receiveDamage(damage) {
+        this.health -= damage;
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super();
```
itstheandre: In the viking you could have used the health and strength from the parent - putting it n the super().
:)

PR #1309, opened 2019-10-22
[PAR-FTWD] Thomas - Vikings 

ThomasAlbrt:
Done

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,111 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+
+    attack() {
+        return this.strength;
+    }
+
+    receiveDamage(damage) {
+        this.health -= damage;
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health, strength);
+        this.name = name;
+    }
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+        super(health, strength);
+        this.name = name;
+    }
+
+    attack() {
+        return this.strength;
+    }
+
+    receiveDamage(damage) {
+        this.health -= damage;
```
receiveDamage(damage){
  super.receiveDamage(damage)
}

starter-code/src/viking.js
```diff
@@ -1,11 +1,111 @@
...
+        this.saxonArmy.push(saxon);
+    }
+
+    vikingAttack() {
+        let randomSaxonNum = Math.floor(Math.random() * this.saxonArmy.length);
+        let randomSaxon = this.saxonArmy[randomSaxonNum];
+        let randomVikingNum = Math.floor(Math.random() * this.vikingArmy.length);
+        let randomViking = this.vikingArmy[randomVikingNum];
+
+        if (randomViking.strength >= randomSaxon.health) {
```
ta-web-paris: Normally here you are supposed to call receiveDamage(damage) before removing the dead solider form the array...
To do that you could hold your function inside a variable (that will at the same time call your function and save the result) and then return this variable at the end.

ta-web-paris:
Good Job! 
"Odin Owns You All!"

PR #1315, opened 2019-10-23
Final version without bonus

ParinaKh:
Hello,

Please let me know if it's all good.

Many thanks
Parina

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,114 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(theDamage) {
+    this.health -= theDamage;
+  }
+}
+
+// var soldier = new Soldier(150, 300);
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (this.health > 0)
+      return `${this.name} has received ${damage} points of damage`;
+    if (this.health === 0) return `${this.name} has died in act of combat`;
+  }
+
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,114 @@
...
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
```
receiveDamage(damage){
  super.receiveDamage(damage)
}

starter-code/src/viking.js
```diff
@@ -1,11 +1,114 @@
...
+    this.saxonArmy.push(saxon);
+  }
+  vikingAttack() {
+    const randDomSaxonIndex = Math.floor(Math.random() * this.saxonArmy.length);
+    const randDomVikingIndex = Math.floor(
+      Math.random() * this.vikingArmy.length
+    );
+    const oneSaxon = this.saxonArmy[randDomSaxonIndex];
+    const oneViking = this.vikingArmy[randDomVikingIndex];
+    if (oneSaxon.health === 0) return this.saxonArmy.pop();
```
ta-web-paris: array.pop() will remove the last element of your array.
But here your dead soldier has a random index, so it's maybe not him who is remove from the array.
To that, either you always chose the last soldier of the array to fight and then remove it with array.pop()
Or you can use array.splice(your random index, 1) so it will only remove your dead soldier

starter-code/src/viking.js
```diff
@@ -1,11 +1,114 @@
...
+    const oneViking = this.vikingArmy[randDomVikingIndex];
+    if (oneViking.health === 0) return this.vikingArmy.pop();
+    return oneViking.receiveDamage(oneSaxon.strength);
+  }
+
+  showStatus() {
+    if (this.saxonArmy.length === 0)
+      return "Vikings have won the war of the century!";
+    if (this.vikingArmy.length === 0)
+      return "Saxons have fought for their lives and survive another day...";
```
ta-web-paris: I think it's red on jasmine because you are not removing well you dead soldier.

PR #1319, opened 2019-10-24
[PAR-FTWD]Camille

Millcat:
vikings done without bonus. Hope I would have time do it this week end ! Could be soo fun :D :D


ta-web-paris:
Great job!! :)

PR #1320, opened 2019-10-24
PAR WEBDEV FT - Marie : Lab Vikings - 

Mawee33:
First try on soldier and viking


ta-web-paris:
Well done ! : ) The solution to the Vikings will be pushed soon : )

PR #1345, opened 2019-10-25
PAR WEBDEV FT 164 [Véro Lab viking done]

vero1308:


starter-code/src/viking.js
```diff
@@ -1,11 +1,51 @@
...
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
 
-// Saxon
-class Saxon {}
+  receiveDamage(damage) {
+    this.health = this.health - damage;
```
ta-web-paris: if you call 
`super.receiveDamage(damage)`
It will call the parents method so you don't have to rewrite the logic : )

ta-web-paris:
I don't know if you missed the War class part or if it's a mistake, either way a solution will be pushed, you can come to us if you have any questions about it.
The rest is good 👌

PR #1361, opened 2019-10-27
vikings [MAD DE] Rebecca

Rebeccawood:
had trouble with the 'war' bit, still figuring out - math.floor(math.random)


ta-web-mad:
Hi Rebecca!

No problem, it took me a while to figure out how to work with the math.random. What's important is that you have undestood how to work with classes and inheritance! Good Job!

Teo :monkey_face: :monkey_face: 

Keep up the good work!

PR #1362, opened 2019-10-27
[PARIS-FTWD-27102019] - Hélène

helenita230:
I started the bonus but couldn't finished it I didn't find how to write the code


ta-web-paris:
Bon travail!! On peut faire le bonus ensemble en renforcement si tu veux :)

PR #1381, opened 2019-10-28
MIA FTWD - October - Ignacio Bellomo

IgnacioBellomo:


ta-web-mia:
Nice job passing all the tests! Hopefully this gives you a better handle on using classes and constructors in object oriented programming. These principles may carry over into other languages if you choose to pursue them! Keep up the good work :)
- Nick

PR #1387, opened 2019-10-28
MIA FTWD October - Daphne Feresin

daph3105:
lab-javascript-vikings


ta-web-mia:
Nice work! This sort of object oriented coding paradigm is becoming the norm across more languages than just JavaScript (Python, PHP, etc.). Feeling comfortable with this sort of structure will help you a lot if you decide to pursue other languages. Keep it up :)
- Nick

PR #1403, opened 2019-11-09
Solved the lab. War never changes.

gha665:


ta-web-mia:
P.s. keep in mind naming convention for the pull requests ;)
Keep up the good work!

PR #1406, opened 2019-11-10
[MIA-PTWD-OCT2019] Humberto & Ashraf

hbuniotto:
Some code on the Bonus section but couldn't make much progress beyond that. Ashraf and I couldn't figure out how have a Viking attack and Saxon (and vice-versa), and how to calculate the damage taken.


ta-web-mia:
Good job on the lab and bonuses. Objects and Classes will become very useful later on as you will see.
Keep up the good work!

PR #1478, opened 2020-01-14
[BER-FT]Shaul Jason

jfrank325:


starter-code/src/viking.js
```diff
@@ -1,11 +1,108 @@
...
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    let vikingHealth = (this.health -= damage);
+    if (vikingHealth > 0) {
+      return `${this.name} has received ${damage} points of damage`;
+    } else if (vikingHealth <= 0) {
```
DalinaLeila: no need to do an `else if`, can directly return your string because it is clear that this will only return if the vikingHealth > 0 is not true on line 24

starter-code/src/viking.js
```diff
@@ -1,11 +1,108 @@
...
-class Saxon {}
+class Saxon extends Soldier {
+  constructor(health, strength) {
+    super(health, strength);
+  }
+  receiveDamage(damage) {
+    let saxonHealth = (this.health -= damage);
+    if (saxonHealth > 0) {
+      return `A Saxon has received ${damage} points of damage`;
+    } else if (saxonHealth <= 0) {
```
DalinaLeila: see comment above :)

starter-code/src/viking.js
```diff
@@ -1,11 +1,108 @@
...
+      return `A Saxon has died in combat`;
+    }
+  }
+}
 
 // War
-class War {}
+class War {
```
DalinaLeila: 💯

PR #1480, opened 2020-01-14
[PAR-WDFT-0120] Pierre Etryhard - ongoing

PierreEtryhard:


Jepassaispar:
Great job, I found out why the ternary operator wasn't working, basically you need to return the full ternary operator for it to work.
instead of : 
```
        if (this.health > 0) {
            return `A Saxon has received ${dmg} points of damage`;
        } else {
        }
```
You need to do this : 
```return (this.health > 0) ? `A Saxon has received ${dmg} points of damage` : `A Saxon has died in combat`;```
It won't work on your last showStatus function as you are declaring 3 different statements, a way to make it work would be : 
```
  showStatus() {
      return(
    this.saxonArmy.length === 0
      ? "Vikings have won the war of the century!"
      : this.vikingArmy.length === 0
      ? "Saxons have fought for their lives and survived another day..."
      : this.saxonArmy.length >= 1 && this.vikingArmy.length >= 1
      ? "Vikings and Saxons are still in the thick of battle."
      : void(0)
      )
  }
```

Anyway all good !

PR #1483, opened 2020-01-14
[BER-FT] Sebastian - Elsa

ElsaFil:


starter-code/src/viking.js
```diff
@@ -1,11 +1,100 @@
...
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (this.health > 0) {
+      return this.name + " has received " + damage + " points of damage";
+    } else {
```
DalinaLeila: no need for the else here

starter-code/src/viking.js
```diff
@@ -1,11 +1,100 @@
...
+      return this.name + " has died in act of combat";
```
DalinaLeila: Works fine, you might want to use backticks and the $ syntax. Will make the code even cleaner :)

starter-code/src/viking.js
```diff
@@ -1,11 +1,100 @@
...
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  receiveDamage(damage) {
+    this.health -= damage;
+
+    if (this.health > 0) {
+      return "A Saxon has received " + damage + " points of damage";
+    } else {
```
DalinaLeila: see the comment above

ElsaFil:
> Looks very good! 👍 👍 Attempt the super Bonus :)

Thanks for the comments! You can see the new changes and also the super bonus 😊💪

PR #1489, opened 2020-01-14
[BER-FT] Laurel & Vaso

lwrocks:
Done

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,72 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(damage) {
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,72 @@
...
+  }
+  vikingAttack() {
+    //let saxonArmy = this.saxonArmy
+    let randomSaxon = this.saxonArmy[
+      Math.floor(Math.random() * this.saxonArmy.length)
+    ];
+    let randomViking = this.vikingArmy[
+      Math.floor(Math.random() * this.vikingArmy.length)
+    ];
+    return randomSaxon.receiveDamage(randomViking.attack());
```
fonsogms: All good in general,
In this last part, you needed to make and if statement when the random saxon is dead, to splice it out of the array.
something like this:
`` if (this.saxonArmy[randomSaxon].health <= 0) {
      this.saxonArmy.splice(randomSaxon, 1);``

PR #1492, opened 2020-01-14
[PAR-WDFT] - Sylvie et Nina - ongoing

Gorlanova:
I don't understand why I get an error in saxonAttack()
(that infects my getStatus() as well since it's pulling the array population from it)

It expects to get a population of 0 in case of a viking attack but viking health is huge and saxon strength very weak, so after one attack (as specified in the exercise) the viking is still alive... Should we make the saxon attack the viking until he dies ? that's not clear in the instructions so I'm not sure...

Thanks !!

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,90 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength){
+        this.health = health;
+        this.strength = strength;
+    };
+
+    attack(){
+        return this.strength;
+    };
+
+    receiveDamage(damage){
+        this.health -= damage;
+    };
+};
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength){
+        super(health, strength);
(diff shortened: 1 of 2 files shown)


Jepassaispar:
First bug, the viking shoud die if the damage is superior or equal to his health, so : 
```
if (damage>=viking.health) {
            this.vikingArmy.splice(vIndex, 1);
        };
```
instead of : 
```
if (damage>viking.health) {
            this.vikingArmy.splice(vIndex, 1);
        };

```
The test should have  throw you an error for the Saxon as well :/

Second bug, the show Status one is just that you are missing the "d" at "survived".

For your question, the 2 armies are fighting each other at each turn until there is no one left in a side or another. 
Whether one of them die or not after one attack, the function will be re-executed with another random Viking and another random Saxon until one of the array is empty.
I don't know if that answers your question, if not feel free to check with me :)

Gorlanova:
Aaaah thanks !!
It absolutely makes sense, thank you !

Dumb Dumb I’ll correct the code right now haha

PR #1493, opened 2020-01-14
[BER-FT-WD-012020] Ciwan / Martin

martinschmal:


starter-code/src/viking.js
```diff
@@ -1,11 +1,101 @@
...
-class Viking {}
+class Viking extends Soldier {
+	constructor(name, health, strength) {
+		super(name, health, strength);
+
+		this.name = name;
+		this.health = health;
+		this.strength = strength;
+	}
+
```
itstheandre: in this viking constructor you did something redundant. when you called for super(name, health, strngth) super is what is coming from the class that the current one is extending.
1st, Soldier doesnt have a name :)
and after that, you redeclared the values with this.health = health and this.strength = strength. By calling the super you dont need to redefine these values :)

PR #1494, opened 2020-01-14
[WD-FT-BERLIN] Alexis Perdikakis

Perdikal:


starter-code/src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+  attack() {
```
fenrew: This method is useless because you wrote "class Viking extends Soldier". This means that your Viking class will inherit the attack() method (and all other methods) from the Soldier class.  ==> remove line 21-23

starter-code/src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+  }
+  battleCry() {
+    return `Odin Owns You All!`;
+  }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  attack() {
```
fenrew: Same applies here

PR #1495, opened 2020-01-14
[PAR][WDFT][012020] Vikings [Tayssir][Yannick] ongoing

taytay81:


TatijanaRajcic:
Hey guys!
Your code is very good!
Some remarks:
- Don't put Capital letters on your placeholders. It does not impact your code but the best practice is using camelCase for variables and placeholders.
- I like the fact that you created your own custom functions getRandomSoldier and removeSoldier. It makes your code even clearer!
Just one thing: this code is repeating itself:
```
    const returnValue = randomSaxon.receiveDamage(randomViking.strength);
    if (randomSaxon.health <= 0)
      this.removeSoldier(randomSaxon, this.saxonArmy);
    return returnValue;
```


```
    const returnValue = randomViking.receiveDamage(randomSaxon.strength);
    if (randomViking.health <= 0)
      this.removeSoldier(randomViking, this.vikingArmy);
    return returnValue;
```

Maybe you could think of a way to refactor it?

Based on your code I wrote this method:


```
  checkSoldierAfterAttack(attackerArmy,victimArmy) {
    const victim = this.getRandomSoldier(victimArmy)
    const attacker = this.getRandomSoldier(attackerArmy)
    const returnValue = victim.receiveDamage(attacker.strength)
    if (victim.health <= 0) 
      this.removeSoldier(victim,victimArmy)
    return returnValue
  }
```

But other solutions are good too! If you come up with an idea, don't hesitate to write me :)

Good job :) :v:

PR #1497, opened 2020-01-14
[PAR-WDFT-0120] - María.MUNIZAGA + Remy.Pham ongoing

mmunizaga:


TatijanaRajcic:
Hey guys! Good job :) Here are some remarks:

- In both your Viking and Saxon classes, your are defining receiveDamage all over again. Instead of writing "this.health -= damage" in both the receiveDamage functions that are inside Viking and Saxon class, you could use super.receiveDamage(damage) and then write the rest of the code that is different from the parent method.
```
// Viking
        super(health, strength);
        this.name = name;
    }
    receiveDamage(damage){
        // this.health -= damage;
        super.receiveDamage(damage)
        if(this.health > 0) return `${this.name} has received ${damage} points of damage`;
    }
    battleCry(){
        return "Odin Owns You All!";
    }
}

// Saxon
class Saxon extends Soldier {
    receiveDamage(damage){
        // this.health -= damage;
        super.receiveDamage(damage)
        if(this.health > 0) return `A Saxon has received ${damage} points of damage`;
    }
}
```

Here, it does not change a lot because the receiveDamage of the parent only contains one line, but if the code was longer, you would have written a lot of repetition code.

- Since you are using Math.floor twice for Viking and twice for Saxon, it would be nice to have dedicated functions. It was not asked so it's just to make your code even cleaner:

```
  getRandomSaxon() {
    return Math.floor(Math.random() * (this.saxonArmy.length - 1));
  }
  getRandomViking() {
    return Math.floor(Math.random() * (this.vikingArmy.length - 1));
  }
```

- Regarding the bonus you did, here are some comments:

 ```
   Attack(whoAtacks){
        let randomViking = this.vikingArmy[Math.floor(Math.random()*(this.vikingArmy.length-1))];
        let randomSaxon = this.saxonArmy[Math.floor(Math.random()*(this.saxonArmy.length-1))];
        let attackSaxon = randomSaxon.receiveDamage(randomViking.strength); // here, you are executing the attackSaxon...
        let attackViking = randomViking.receiveDamage(randomSaxon.strength); // ..and the vikingAttack.
        // therefore, your parameter is of no use.
        
        // I guess that what you are trying to do with the following is actually simulating an attack depending on the parameter
        // that it's exactly why you can't write both attackSaxon and attackViking in the previous lines, because you want to call them depending on your parameter
        whoAtacks === viking? attackSaxon: attackViking;

        attackSaxon === `A Saxon has died in combat`? this.saxonArmy.splice(randomSaxon):'';
        attackViking === `${randomViking.name} has died in act of combat`? this.vikingArmy.splice(randomViking):'';

    }
```
I modified it a bit and here is my proposition:

```
    attack(whoAtacks){
        let vikingIndex = Math.floor(Math.random()*(this.vikingArmy.length-1))
        let randomViking = this.vikingArmy[vikingIndex];
        let saxonIndex = Math.floor(Math.random()*(this.saxonArmy.length-1));
        let randomSaxon = this.saxonArmy[saxonIndex];
        if (whoAtacks === "viking") {
            let result = randomSaxon.receiveDamage(randomViking.strength) 
            if (result === `A Saxon has died in combat`) this.saxonArmy.splice(saxonIndex,1)
        }
        else {
            let result = randomViking.receiveDamage(randomSaxon.strength);
            if (result === `${randomViking.name} has died in act of combat`) this.vikingArmy.splice(vikingIndex,1)
        } 
    }
```
Then, if your object is called war (using var war = new War() ), you'll simulate a viking attack using war.attack("viking") . If you use any other parameter, it will trigger a saxon attack.

Great job :)

PR #1499, opened 2020-01-15
[BER-FT-0120] Johnny and Akua

JohnnyAL:


starter-code/src/viking.js
```diff
@@ -1,11 +1,76 @@
...
 
 // War
-class War {}
+class War {
+  constructor() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+  }
+
+  addViking(Viking) {
```
fenrew: let the arguments be written in only lower case letters as it can be a bit confusing when the first letter is capitalized (that is normally only used when defining a class, constructor, models, etc)

starter-code/src/viking.js
```diff
@@ -1,11 +1,76 @@
...
+    this.saxonArmy.push(Saxon);
+  }
+
+  vikingAttack() {
+    let randomViking = Math.floor(Math.random() * this.vikingArmy.length);
+    let randomSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+    let result = this.saxonArmy[randomSaxon].receiveDamage(
+      this.vikingArmy[randomViking].attack()
+    );
+    if (this.saxonArmy[randomSaxon].health < 0) {
```
fenrew: A small bug in this if statement: If health == 0 it will technically be dead, but remain in the saxonArmy. health <= 0 will solve the issue

PR #1520, opened 2020-01-18
[PAR WDFT] Mehmet 

MehmetCan75:


starter-code/src/viking.js
```diff
@@ -1,11 +1,48 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+        Soldier.prototype.attack = function attack() {return this.strength}
+        Soldier.prototype.receiveDamage = function(damage) {this.health -= damage}
```
ta-web-paris: Using the class syntax, this isn't how you are supposed to declare the class methods, the constructor is for the class properties, and then you declare it's methods within the class as follows : 

```
class Soldier {
  constructor(health, strength) {
    this.health = health;
    this.strength = strength;
  }

  attack() {
    return this.strength;
  }
  receiveDamage(damage) {
    this.health -= damage;
  }
}
```

starter-code/src/viking.js
```diff
@@ -1,11 +1,48 @@
...
+
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health,strength);
+        this.name = name;
+        Viking.prototype.receiveDamage = function receiveDamage(damage) {
+            this.health -= damage;
```
ta-web-paris: Here you could have called the parents function by using, super.receiveDamage(damage)
For this case it's not a problem since there is only 1 line of code, but if there was more logic to the function you
would be writing it over again.

ta-web-paris:
A little confusion with the constructor function syntax and the class syntax, but the implementation of each methods work, so well done : ) 
Closing the PR.

PR #1550, opened 2020-01-20
[PAR-FT-012020] Yuval BUGAJSKI ongoing

YuvalBu:


TatijanaRajcic:
Hey Youval! Nice work on the first three classes :) 

Some of my advice:

1) In your Viking and Saxon classes, you could use the super property, to inherit from the parent the receiveDamage function. To summarize, you can have 3 cases:
- A subclass inherit from its parent class the exist same method and you don't want to change it: then, you don't have to write it at all in the subclass. Ex:"attack" exists in the parent class, and should be used in the exact same way in the subclasses, so that's why you are not defining them again in the subclasses
- A subclass has a method with the same name as one in the parent class but is completely different from the parent one. Then, you have to define it from zero in the subclass. This will overide the parent method with the same name
- A subclass has a method with the same name as one in the parent, and it should inherit from it but it completes it. Ex: in Viking and Saxon classes receiveDamage do AT LEAST the exact same thing as the parent method with the same name, and they do specific things too. Here, instead of writing "`this.health -= theDemage`" again in both subclasses, you could use super:

```
        super(health, strength);
        this.name = name;
    }
    receiveDamage(theDemage){
// HERE
        super.receiveDamage(theDemage)
        if (this.health > 0){
            return(`${this.name} has received ${theDemage} points of damage`);
            }
        else {
            return(`${this.name} has died in act of combat`);
            }
        }

    battleCry(){
        return "Odin Owns You All!";
    }
}
```

In this case, it does not change a lot because the parent receiveDamage is super short, but it could be really useful if it was longer!

2) Have you tried to do the vikingAttack and saxonAttack? The main idea in both methods is to randomly generate a soldier, on one of them use the method Attack and on the other Receive Damage.

Don't hesitate to try it or come and see me if you want  :))

PR #1554, opened 2020-01-24
[AMS-PT-012020] - Jonas B.

jonasbarre:


starter-code/src/viking.js
```diff
@@ -1,11 +1,87 @@
...
+        const vikingSoldierSelector = Math.floor(Math.random()*this.vikingArmy.length)
+        const vikingSoldier = this.vikingArmy[vikingSoldierSelector]
+        const result = vikingSoldier.receiveDamage(saxonSoldier.strength)
+        if (vikingSoldier.health < 1) {
+            this.vikingArmy.splice(vikingSoldierSelector, 1)
+        }
+        return result
+    }
+    showStatus(){
+
```
LloydC: if (this.saxonArmy.length === 0) 
        return "Vikings have won the war of the century!";
    else if (this.vikingArmy.length === 0) 
        return "Saxons have fought for their lives and survive another day...";
    else if (this.saxonArmy.length >= 1 && this.vikingArmy.length >= 1) 
  }

PR #1555, opened 2020-01-24
[AMS-PT-012020]-Rajmohan B - Lab war games completed

Rajmohankct2003:


starter-code/src/viking.js
```diff
@@ -1,11 +1,91 @@
...
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health , strength) {
+        super(health, strength);
+        this.name = name;
+    }
+    receiveDamage(damage) {
+        this.health = this.health - damage;
+        if (this.health > 0 ) {
```
LloydC: Also can be written like return this.health > 0 ? `${this.name} has received ${damage} points of damage` :  `${this.name} has died in act of combat`

PR #1556, opened 2020-01-25
done

fhoman:


starter-code/src/viking.js
```diff
@@ -1,11 +1,176 @@
...
+
+    addSaxon(saxon) {
+
+        this.saxonArmy.push(saxon);
+
+    }
+    vikingAttack() {
+
+
+
```
LloydC: Also can be return this.vikingArmy.length > 0 ? this.attack(this.vikingArmy, this.saxonArmy) : `There is no Viking army to attack with`

starter-code/src/viking.js
```diff
@@ -1,11 +1,176 @@
...
+        } else {
+
+            return `There is no Viking army to attack with`
+
+        }
+    }
+
+    saxonAttack() {
+
```
LloydC: Also can be return this.saxonArmy.length > 0 ? this.attack(this.saxonArmy, this.vikingArmy) : `There is no Saxon army to attack with`

PR #1560, opened 2020-01-26
BoyuanShi_PTWD_AMS_2020Jan26

sbyshjd:


starter-code/src/viking.js
```diff
@@ -1,11 +1,109 @@
...
+        this.vikingArmy = [];
+        this.saxonArmy = [];
+    }
+    addViking(viking) {
+        this.vikingArmy.push(viking);
+    }
+    addSaxon(saxon) {
+        this.saxonArmy.push(saxon);
+    }
+    attack(a) {
```
LloydC: I think it works better if you do like attack( attackingArmy, defendingArmy) then instead of those two "if" statements you can just use the "attackingArmy" and "defendingArmy" parameters accordingly.

starter-code/src/viking.js
```diff
@@ -1,11 +1,109 @@
...
+        return 'Odin Owns You All!'
+    }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    receiveDamage(damage) {
+        this.health-=damage;
+        if(this.health > 0) {
```
LloydC: Also can be return this.health > 0 ? `A Saxon has received ${damage} points of damage` :  `A Saxon has died in combat`

starter-code/src/viking.js
```diff
@@ -1,11 +1,109 @@
...
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor (name,health,strength) {
+        super(health,strength);
+        this.name = name;
+    }
+    receiveDamage(damage) {
+        this.health-=damage;
+        if(this.health > 0) {
```
LloydC: Also can be return this.health > 0 ? `${this.name} has received ${damage} points of damage` : `${this.name} has died in act of combat`

PR #1571, opened 2020-02-15
BERLIN WD PT -  Ian

ian-kerins:


starter-code/src/viking.js
```diff
@@ -1,11 +1,148 @@
...
+        }
+        else if(this.saxonArmy.length == 0){
+            return 'Vikings have won the war of the century!';
+        }
+        else{
+            return 'Vikings and Saxons are still in the thick of battle.';
+        }
+    }
+
+}
```
numbing: nice job, all good :)

PR #1579, opened 2020-02-15
BER WDPT2020 Solange

solangeohana:


starter-code/src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+
+    addSaxon(saxon){
+        this.saxonArmy.push(saxon)
+    }
+
+    vikingAttack(){
+     let randomSaxonPosition = Math.floor((Math.random() * this.saxonArmy.length))
+     let randomVikingPosition = Math.floor((Math.random() * this.vikingArmy.length))
+
+        this.saxonArmy[randomSaxonPosition].receiveDamage(this.vikingArmy[randomVikingPosition].strength)
```
numbing: here you should write 
let result = this.saxonArmy[randomSaxonPosition].receiveDamage(this.vikingArmy[randomVikingPosition].strength)

starter-code/src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+
+        if(this.saxonArmy[randomSaxonPosition].health <= 0){
+            this.saxonArmy.splice(randomSaxonPosition, 1)
+        }
+    //return ???
```
numbing: and here:
 return result

PR #1582, opened 2020-02-15
BER-WDPT-202002-Sonja L. & Sjard

Asphalthexe:
BER-WDPT-Sonja L. & Sjard

finish

### DIFF
--- a/starter-code/src/viking.js
+++ b/starter-code/src/viking.js
@@ -1,11 +1,161 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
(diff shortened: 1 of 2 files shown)


starter-code/src/viking.js
```diff
@@ -1,11 +1,161 @@
...
+        //static
+        vikingArmy = saxArray
+        saxonArmy = vikArray
+    }
+
+
+}
+
+*/
+// let battlefield = new War(saxArray[],)
```
numbing: its all good, nice job but its commented :D

PR #1584, opened 2020-02-15
Loriel WDPT Berlin

Loriel-S:
Thanks :)


starter-code/src/viking.js
```diff
@@ -1,11 +1,61 @@
...
+        return `A Saxon has died in act of combat`
+}
+}
+
+}
+
 
 // War
-class War {}
+class War {}
```
numbing: War is missing, if you want i can explain you how it works and what you should do but ask me about it please

PR #1585, opened 2020-02-16
BER - WDPT - Pia and Francesco

pgmdf:


starter-code/src/viking.js
```diff
@@ -1,11 +1,151 @@
...
+        else if (this.saxonArmy.length <= 0) {
+            return `Vikings have won the war of the century!`
+        }
+        else if (this.vikingArmy.length <= 0) {
+            return `Saxons have fought for their lives and survived another day...`
+        }
+    }
+
+}
+
```
numbing: its all good, nice job

PR #1604, opened 2020-02-24
AMS FT FEB - Alexandra Muskovszky

Carriettta:


MFn-Scarboy:
- Technical Requirements (10/10)
- Creative Code (9/10)
- Code Quality (10/10)

Very nice ! All the bonus done.
Literally the only thing I would change would be to add line breaks between functions to make it a bit more readable but that's personal preference and your code is very readable as is.

Also, try to start commenting your code, the idea behind this is that future colleagues and employers have an easier time understanding your code, this is a big plus for them. So in the future if you have a big function or something important adding a comment with a small summary of what it does is very good practice.

PR #1660, opened 2020-03-31
[PAR WDFT 0320] William Hermozo ongoing

williamhzo:


ta-web-paris:
Hey William, great job !

For the ternary, have you tried writing "return" just before it ? the function expects a return value
For the war part, try to create an index value for your math random, because when you call it later to remove the viking / saxon, it does a math random again (and that could create another value)

Otherwise excellent

Hit me up for questions

Nina

PR #1665, opened 2020-04-01
BER-FTWD-Sami

Alba-Tros:


src/viking.js
```diff
@@ -1,11 +1,81 @@
...
+        this.vikingArmy.push(Viking);
+    }
+    addSaxon(Saxon) {
+        this.saxonArmy.push(Saxon);
+    }
+    vikingAttack() {
+        let saxonRandom = Math.floor(Math.random() * this.saxonArmy.length);
+        let vikingRandom = Math.floor(Math.random() * this.vikingArmy.length);
+
+        this.Saxon[saxonRandom].receiveDamage(
```
fenrew, line 60: I dont think this like is correct (this.Saxon).

this.Saxon refers to a variable called this.Saxon located in the constructor of the War class.
Instead I believe it should say: this.saxonArmy[saxonRandom] because all your saxons are within this array:)

PR #1680, opened 2020-04-03
PAR WDFT 202003 - Anne Patana - Week2 Day2

apatana:


ta-web-paris:
Hi Anne !

Great job !
Careful about your SaxonAttack function, it's not exactly the same as the Viking attack (you need to reverse the variables for it to go the other way)

Otherwise very nice :)

Nina

PR #1681, opened 2020-04-03
PAR-FT-032020 Yuval Bugajski 

YuvalBu:


ta-web-paris:
Great job Yuval !!
Only point : you don't have to rewrite the constructor if you only copy it from Soldier, you need to rewrite it and use super only if you add something (like the name for the Viking)

Nina

PR #1683, opened 2020-04-03
[PAR WDFT] Vladimir Leloup

Vlp90:
exercice done without bonus

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,131 @@
 // Soldier
-class Soldier {}
+class Soldier {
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hey Vladimir ! 

Quick points :

- you don't need to repeat the function "attack" since it's already defined in Soldier (and you copied that class)

- same for Saxon, you just need to write the receiveDamage since it's the only difference !

Otherwise very nice !!

Hit me up if you have questions :) 

Nina

PR #1685, opened 2020-04-03
[BCN-FT-032020] Hélène Stephan

HeleneStephan:


tawebbcn:
Very clean code and all iterations done except bonus. Keep up the good work!

PR #1687, opened 2020-04-04
done

alpagor:
Primer lab completado 100%!!!! :"D

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,88 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+  receiveDamage(damage) {
+    this.health -= damage;
+  }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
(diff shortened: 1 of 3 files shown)


src/viking.js
```diff
@@ -1,11 +1,88 @@
...
+    this.saxonArmy = [];
+  }
+  addViking(Viking) {
+    this.vikingArmy.push(Viking);
+  }
+  addSaxon(Saxon) {
+    this.saxonArmy.push(Saxon);
+  }
+  vikingAttack() {
+    var oldHealth = this.saxonArmy[0].receiveDamage(
```
tawebbcn, line 64: The repo instruction was to do a random viking/saxon.

src/viking.js
```diff
@@ -1,11 +1,88 @@
...
+      this.vikingArmy[0].strength
+    );
+    if (this.saxonArmy[0].health <= 0) {
+      this.saxonArmy.length -= 1;
```
tawebbcn, line 68: This will not remove an element from the array. See methods .splice(), .pop(), .shift() that would do that instead.

tawebbcn:
Excellent work!! You managed to finish all lab so congrats! I would only recommend you to check the two specific comments we mentioned, specially the one regarding removing elements from an array.
Also, please remember to change the title of your pull request according to the required format.

PR #1688, opened 2020-04-04
BCN-FT-032020 Siiri S

fetaplop:
Sorry I didn't know how to not send Alberto's as well but mine is viking.js. Then again maybe you see something in comparison that I don't see bc Alberto got his working with jasmine, I didn't and they look pretty similar to me. (really wanted to do the bonus but jasmine kinda ate my motivation)


src/viking.js
```diff
@@ -1,11 +1,52 @@
...
+    super(health, strength);
+    this.name = name;
+  }
+  // does this work???????????
+  receiveDamage(damage) {
+    this.health = this.health - damage;
+    if (this.health <= 0) {
+      return `${this.name} has died in act of combat`;
+    } else {
+      return `${this.name} has received ${this.damage} points of damage`; //NAMEDAMAGE
```
tawebbcn, line 27: here you're referring to `this.damage` as if it was a property, but it's a parameter of the receiveDamage() method, so it should be used without the "this." 
`return '${this.name} has received ${damage} points of damage'; `

src/viking.js
```diff
@@ -1,11 +1,52 @@
...
+  battleCry() {
+    return "Valhallaaaaaaaa!"; // not gonna use the given one, this is better (and like jasmine doesnt work anyhow so..)
+  }
+}
+//Viking.attack(this.strength); // added this bc Alberto told me it might help????????
 // Saxon
-class Saxon {}
-
+class Saxon extends Soldier {
+    super(health, strength);
```
tawebbcn, line 37: Here you forgot to create a constructor, in order to use super() and inherit from the Soldier class properties... this is what breaks jasmine tests :(

PR #1689, opened 2020-04-04
lab-javascript-vikings

RicardVillalba:
Hi, I'm working on the bonus part I'll push it again after finish the iteration 4.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -50,4 +50,19 @@ class Saxon extends Soldier {
 }
 
 // War
-class War {}
+class War {
+  vikingArmy = [];
+  saxonArmy = [];
+
+  addViking(Vikingobject) {
+    this.vikingArmy.push(Vikingobject);
+  }
+
+  addSaxon(Saxonobject) {
+    this.saxonArmy.push(Saxonobject);
+  }
+
+  vikingAttack() {}
+
+  saxonAttack() {}
+}
--- a/src/viking.js
+++ b/src/viking.js
@@ -50,6 +50,4 @@ class Saxon extends Soldier {
 }
 
 // War
-class War {
-  addViking() {}
-}
+class War {}
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,55 @@
 // Soldier
-class Soldier {}
-
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,68 @@
...
+  }
+  receiveDamage(damage) {
+    this.health = this.health - damage;
+  }
+}
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(name);
```
tawebbcn, line 17: The super keyword is used to use parameters from the parent class, so in this case it is the parameters health and strength the ones that should be passed from the parent Soldier. Not the parameter name.

src/viking.js
```diff
@@ -1,11 +1,68 @@
...
+    this.name = name;
+    this.health = health;
```
tawebbcn, line 19: Same as before comments, variables health and strength should not be declared as they are passed from parent.

src/viking.js
```diff
@@ -1,11 +1,68 @@
...
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(damage) {
+    this.health = this.health - damage;
+    if (this.health > 0) {
+      return this.name + " has received " + damage + " points of damage";
```
tawebbcn, line 29: Try to do string interpolation for this kind of strings. They are better for readability.

src/viking.js
```diff
@@ -1,11 +1,68 @@
...
+  }
+
+  battleCry() {
+    return "Odin Owns You All!";
+  }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
```
tawebbcn, line 41: You forgot to use super inside constructor to create health and strength properties thast are super from parent Soldier.

src/viking.js
```diff
@@ -1,11 +1,68 @@
...
+    } else if (this.health <= 0) {
+      return "A Saxon has died in combat";
+    }
+  }
+}
 
 // War
-class War {}
+class War {
+  vikingArmy = [];
```
tawebbcn, line 54: The two arrays should be declared inside a constructor. Same as previous classes.

tawebbcn:
Very good job on the Lab. Just remember that super is used to get properties from parent class and not to declare new ones. Good that you started the bonus part! Keep up the good work.

PR #1692, opened 2020-04-04
[BCN-FT-032020] Daniel Hamekasi

dhamekasi:


src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+  addViking(viking) {
+    this.vikingArmy.push(viking);
+  }
+
+  addSaxon(saxon) {
+    this.saxonArmy.push(saxon);
+  }
+
+  vikingAttack() {
+    let result = this.saxonArmy[this.saxonArmy.length - 1].receiveDamage(
```
tawebbcn, line 71: For better readability, we would recommend to store variables like the index  or even the current viking/saxon outside before this in order to avoid long methods that are harder to read.

tawebbcn:
Other than the viking/saxon needed to be random as discussed in class, the Lab solution was perfect! We would only recommend to avoid long lines of code that are harder to read by declaring more variables used in them.

PR #1693, opened 2020-04-04
done + bonuses - [WDFT-BCN-ALEX.A]

skanndar:


src/viking.js
```diff
@@ -1,11 +1,96 @@
...
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+  receiveDamage(theDamage) {
+    this.theDamage = theDamage;
```
tawebbcn, line 22: Not really necessary to add theDamage as a property of the class Viking

src/viking.js
```diff
@@ -1,11 +1,96 @@
...
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  constructor(health, strength) {
+    super(health, strength);
+  }
+  receiveDamage(theDamage) {
+    this.theDamage = theDamage;
```
tawebbcn, line 41: same here, you don't need to add theDamage as a property of Saxon, you can use it directly as a parameter:
`
`A Saxon has received ${theDamage} points of damage`;
`

src/viking.js
```diff
@@ -1,11 +1,96 @@
...
+  vikingAttack() {
+    let result = this.saxonArmy[this.saxonArmy.length - 1].receiveDamage(
+      this.vikingArmy[this.vikingArmy.length - 1].strength
+    );
+
+    if (this.saxonArmy[this.saxonArmy.length - 1].health <= 0) {
+      this.saxonArmy.pop();
+    }
+    return result;
+  }
```
tawebbcn, line 74: Works well, but in that case it's always the last viking of the vikingArmy array attacking the last saxon of the saxonArmy array 😉 You could try choosing a saxon and viking at random like specified in the lab instructions (although the jasmine tests were not checking it)

PR #1695, opened 2020-04-05
[BCN-FT-032020-Maxime Lagrange]

Gitrader:
Mandatories iterations done, working on the bonus

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,70 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,70 @@
...
+}
 
 // War
-class War {}
+class War {
+  constructor() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+  }
+  addViking(Viking) {
```
tawebbcn, line 64: here by convention, you should write "viking" with the first letter lowercase (it's not exactly the Viking constructor you're passing as a parameter, but an instance of it)

src/viking.js
```diff
@@ -1,11 +1,70 @@
...
+    this.vikingArmy.push(Viking);
+  }
+  addSaxon(Saxon) {
```
tawebbcn, line 67: same here ;)
otherwise, everything good!

PR #1696, opened 2020-04-05
[BCN-FT-032020] Ariadna & Alberto

arimagic-8bit:
Hello! 

The lab is all done. 

Thanks! :)

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,60 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+  }
+}
+
+Soldier.receiveDamage(5);
 
 // Viking
-class Viking {}
+
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,92 @@
...
+  }
+  vikingAttack() {
+    var saxonNewHealth = this.saxonArmy[0].receiveDamage(
+      this.vikingArmy[0].strength
+    );
+    if (this.saxonArmy[0].health <= 0) {
+      this.saxonArmy.length -= 1;
+    }
+    return saxonNewHealth;
+  }
```
tawebbcn, line 73: Works well, but in that case it's always the first viking of the vikingArmy array attacking the first saxon of the saxonArmy array 😉 And then you remove the last saxon from the saxonArray if the first one is dead.
To be more precise, you could try choosing a saxon and viking at random like specified in the lab instructions, and removing that exact random saxon in case he's dead (although the jasmine tests were not checking it)

src/viking.js
```diff
@@ -1,11 +1,92 @@
...
+  saxonAttack() {
+    var vikingNewHealth = this.vikingArmy[0].receiveDamage(
+      this.saxonArmy[0].strength
+    );
+    if (this.vikingArmy[0].health <= 0) {
+      this.vikingArmy.length -= 1;
+    }
+    return vikingNewHealth;
+  }
```
tawebbcn, line 82: Same comment here 😉

PR #1697, opened 2020-04-06
PAR WDFT MARCH 2020 SEBASTIEN DUBOIS

sdubois90:


ta-web-paris:
Hey Sebastien !

Quick points : 
- your class War : it's better to use a constructor here since it's a class : 
>> constructor(){
        this.vikingArmy = [];
        this.saxonArmy = [];
    };

Otherwise you got the point, nice job :)

sdubois90:
Thanks a lot for your explanation!

Have a good weekend,
Sébastien

----- Mail original -----
De: "TA paris" <notifications@github.com>
À: "ironhack-labs/lab-javascript-vikings" <lab-javascript-vikings@noreply.github.com>
Cc: "sdubois90" <seb.dubois15@free.fr>, "Author" <author@noreply.github.com>
Envoyé: Lundi 6 Avril 2020 10:19:32
Objet: Re: [ironhack-labs/lab-javascript-vikings] PAR WDFT MARCH 2020 SEBASTIEN DUBOIS (#1697)





Hey Sebastien ! 

Quick points : 

    * your class Viking : "super" is what you copy from the Soldier class (so only health and strength) and you don't need to define them after (since they're copied) but you don't need to put name in super (Soldier didn't have a name) and then you need to define it after (because it's new, and you did it so that's good !) 
    * your class War : it's better to use a constructor here since it's a class : 






constructor(){ 
this.vikingArmy = []; 
this.saxonArmy = []; 
}; 

Otherwise you got the point, nice job :)

PR #1698, opened 2020-04-06
Louis Alibert - PAR - done

lalib19:


ta-web-paris:
Hey Louis !

Quick points :

- you don't need to repeat the "attack" function since it's the same as Soldier (you created a copy so it's already there even if you don't see it) you only need to write what's different, that's why classes are helpful ! 

- same for Saxon, you only need to write the receive damage function (it's the only difference)

Otherwise very good !

Nina

PR #1828, opened 2020-06-16
[PAR-FT-0620] Menhaj Safi

Menhaj01:


TatijanaRajcic:
Hello Menhaj! Congrats on your lab, it's almost perfect!
The only thing you should keep in mind is that if a parent class has a method that we want the child class to have but to be slightly different, you can build on that, you don't have to rewrite the whole thing: see this [article](https://javascript.info/class-inheritance#overriding-a-method).
In your Viking and Saxon classes, you can replace the line `this.health = this.health - damage;` by     `super.receiveDamage(damage);`
Great work!

PR #1836, opened 2020-06-24
[REMOTE-WDPT-0602] - Sunitha Mannam 

SunithaMannam:
lab complete


valmgisbert:
Very thorough in using a function also for `health` even if it wasn't in the instructions. Also good on you for reusing the possible elements. Great work!

PR #1837, opened 2020-06-24
[REMOTE-WDPT-0602] - Matthew Howard

mattalanhoward:


valmgisbert:
You can use shorthand operators too in `receiveDamage()` to clean your function; also, you could call it via `super` in your extended classes. In your `viking+saxonAttack()` functions, you can also reuse the `attack()` for each of the classes in your `result`. Great work!

PR #1838, opened 2020-06-24
[WDPT-JUNE/2020] - Willem Prins

willemprins93:
Finish all but parts of bonus

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,104 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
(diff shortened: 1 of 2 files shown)


valmgisbert:
You can use shorthand operators too in `receiveDamage()` to clean your function; also, you could call it via `super` in your extended classes. In your `viking+saxonAttack()`, you should focus on the random soldiers that are in the middle of the attack to have a more thorough result of their health and if they need to be removed or not from the armies (also to stay consistent within your function!). Your `showStatus()` might not be working because of the conditionals; check the amount of equals you're using. Bonuses done. Great work!

PR #1839, opened 2020-06-24
[REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:


valmgisbert:
You can use shorthand operators too in `receiveDamage()` to clean your function; also, you could call it via `super` in your extended classes. Your  `attack(attacker)` is basically the previous two, so you are repeating code here. Excellent work!

PR #1840, opened 2020-06-24
[REMOTE-WDPT-0602] -Nuno Pacheco

nuno-pacheco:
This lab was a pleasure, I will try to do the bonus later.


valmgisbert:
You can use shorthand operators too in `receiveDamage()` to clean your function; also, you could call it via `super` in your extended classes. I'm glad you liked this lab! Please try to do the bonus whenever you can. Very nice job!

PR #1842, opened 2020-06-24
[REMOTE-WDPT-0602] -Rodrigo Hernández

Rodrigohernandezsan:
Lab completed! Some difficulties in the Bonus.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,77 @@
 // Soldier
-class Soldier {}
-
+class Soldier {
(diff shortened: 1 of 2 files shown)


valmgisbert:
Remember to work on spacing for an easier to read code. Good on reusing your functions! you could also `return` the messages instead of inserting them in a variable. In your `viking+saxonAttack()` functions, you can also reuse the `attack()` for each of the classes in your `result`. Great work!

PR #1843, opened 2020-06-25
 [REMOTE-WDPT-0602] Pedro Lamy

PedroMMLamy:


valmgisbert:
you could call your `receiveDamage()` via `super` in your extended classes. Whenever you can, please work on the bonuses. Nice job!

PR #1844, opened 2020-06-25
[REMOTE-WDPT-0602] - Michel Ribbens

AmanoLX:


valmgisbert:
you could call your `receiveDamage()` via `super` in your extended classes. Also use your `attack()` in your `saxon+vikingAttack()`. Your `whoIsAttacking()` is basically repeating the code from the previous two functions. Great work!

PR #1845, opened 2020-06-25
[REMOTE WDPT 0620] Lynn W-B

lynn-w-b:
Done - Jasmine still failing one test!

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,108 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength){
+        this.health = health;
+        this.strength = strength;
+    };
+    attack(){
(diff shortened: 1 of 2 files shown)


valmgisbert:
Very nice! You can also try shorthand operators too in `receiveDamage()` to clean your function; also, you could call it via `super` in your extended classes. In your `viking+saxonAttack()` functions, you can also reuse the `attack()` for each of the classes in your `attackDamages`. If you have any failing tests, you should comment a bit either the name or what it looks for, so that I can guide you a bit. From a glimpse, it could be that the instructions ask you to return the result of `receiveDamage()` & `attack()` interaction in your `viking+saxonAttack()` functions, and you are returning something else. Great job!

PR #1847, opened 2020-06-25
[REMOTE-WDPT-0602] Cristina Matamoro

cris-developer:
Unfortunately, i had not time enought to finish it. Done iteration 1,2,3 and half of 4. I will get back to try to finish it

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,62 @@
-// Soldier
-class Soldier {}
+// Soldier iteration 1
+class Soldier {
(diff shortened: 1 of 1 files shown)


valmgisbert:
Be careful with the typo in shorthand op in your first `receiveDamage()`; also, you could call `receiveDamage()` via `super` in your extended classes. You're doing pretty well! Whenever you can, try to finish the bonuses of this lab. Nice work!

PR #1848, opened 2020-06-25
[REMOTE-WDPT-0602] Andre Bacalhau

AndrewMastercode:


valmgisbert:
Be careful with spacing in your code to make it legible. You can use shorthand operators too in `receiveDamage()` to clean your function; also, you could call it via `super` in your extended classes. Also, try using string interpolation in your `receiveDamage()` results. In your `viking+saxonAttack()` functions, you can also reuse the `attack()` for each of the classes in your `damage`. Very nice work!

PR #1849, opened 2020-06-27
[REMOTE-WDPT-0602] - ALEJANDRO GONZALEZ

alejanglez:
Hey!
The IT4 is not working, it broke all the tests..I will try to finf what is happening and finish de IT5.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,101 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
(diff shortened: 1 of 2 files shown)


valmgisbert:
Be careful with spacing in your code to make it legible. You can call `receiveDamage()`  via `super` in your extended classes. In your `viking+saxonAttack()` functions, you can also reuse the `attack()` for each of the classes in your `resultDamage`. The possible issue is that you have `showStatus()` without any code inside, and that breaks Jasmine. Whenever you can, please work on the superbonus. Very nice work!

PR #1873, opened 2020-07-13
[WDFT-REMOTE-2007] Alejandro Hernández

alejandrohj:
Done


IH-WebDev-TA-Remote:
Looks perfect, clever looking for the "died" word to remove soldiers. Although checking if their health would be 0 or less would work as well.
Keep it up!

PR #1874, opened 2020-07-13
[WDFT-REM-2007] BRITTA HELM

brittahelm:
seemed fairly doable, just not completely certain on the bonus (general attack) function since i do tend to forget commas without jasmine's help


IH-WebDev-TA-Remote:
Looks perfect, very good job!
Remember as best practices to use triple equal instead of double equal.

PR #1876, opened 2020-07-13
[WEB-REM-2007] Lanette Lanting

llanting:
Done with the Vikingslab, now we only have to wait until AC: Valhalla drops to see it in action. ;)


IH-WebDev-TA-Remote:
Amazing work! And also the Super Bonus! Keep up the great work!
_Nothing is real. Everything is permitted ;)_

PR #1877, opened 2020-07-13
[WEB-REM-2007] RAFAEL DE LAZARO

rlaz133:
I think I understand these concepts well but still need to double check the syntax every so often.


IH-WebDev-TA-Remote:
Great work! everything looks good and tests are passing! Keep it up!
Syntax is hard because it is new topics but with lots of practice you can polish that knowledge.

PR #1878, opened 2020-07-13
WEB-REM-2007 Bodei

BodeiBrouwer:
It all went really quickly and easily until the bonus, but with a bunch of help from Manish I managed in the end.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,100 @@
 // Soldier
-class Soldier {}
+class Soldier {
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Great work on the lab! Code looks good and all tests passing.
It is ok to ask for some help and hints on how solve the labs. The important part is that you understand everything that is happening in the code and you can then reapply it :)
Keep it up!

PR #1879, opened 2020-07-13
WEB remote 2007 - katie stamey

kmstamey:


IH-WebDev-TA-Remote:
Good job on the lab! but don't be afraid to try those last bonus.

PR #1880, opened 2020-07-13
[WDFT-REM-2007] ALVARO SANCHEZ LAMADRID

AlvaroSLam:
Exercise done, found some resistance with the bonus part, but everything else is done.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,86 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health,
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very good job on the lab! and good try on the bonus. You were missing just a bit to finish it.
1. Returning the result variable 
2. Removing dead vikings/saxons (you can do that by checking if the health of the randomViking is 0 or less and then using a splice. For the index to be spliced you can save the random number in a variable when getting the viking object)

PR #1881, opened 2020-07-13
WEB Development - Remote - July2020 Joel Roca Martinez vikingLAB

JoelRocaMartinez:
I have really enjoyed doing it. Can´t wait to see how to complete the parts that I was missing to do in the LAB!

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,83 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+    }
+    attack() {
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very good job on the lab, seems you confusion was with how to use the receiveDamage method. Because it is a method from the Viking or Soldier class you would need to call it from those objects. Something like randomSaxon.receiveDamage(this.randomViking.strength)
But also you are having another issue, the selected random viking and random saxon should be inside the attack methods so you have a new one everytime the method is called. But that one is more videogame logic than javascrip logic haha.
Very good job overall, keep it up!

PR #1882, opened 2020-07-13
[WDFT-REM-2007] Maëva Merrou

maevamerrou:
Hey, the bonuses are not validated by Jasmine. I don't manage to find out why.
Please help! :)

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,86 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength){
+        this.health = health;
+        this.strength = strength;
+    }
+        attack(){
+            return this.strength;
+    }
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very good job on the lab! I see most is done expect some things breaking in the bonus.
1. First, the return on line 64 is preventing the rest of the code from running in that method. Remember the return is the last thing a function will run.
2. the health comparison the in the condition should be lesser than and not equal, is the result can be negative numbers.
3. Using delete deletes the object, but it might leave an undefined in it's place of the array. In order to correctly remove an item from an array you can use .splice.
4. You want to run the receiveDamage method before the conditional but return it after. To do that just save the method call in a variable and return the variable.

PR #1883, opened 2020-07-13
[WEB - REM - 2007] Marta Lourido 

MartaLourido:
Done! This lab was fun.

I haven't got the last bonus working yet, but I've got the others, after being stuck for a while It finally works.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,119 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor (health, strength) {
+        this.health = health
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very good job on the lab!! Almost all the tests passing. Issues you were having in the last one were:
1. It should be a method of the War class, so it should be inside, similar to the other methods.
2. The return messages should be exchanged
3. The last condition should also return another string :)
But with those 3 small fixes all tests pass. So again, very good job!

PR #1884, opened 2020-07-13
viking lab

GunnerAg:
I still have not found some issues with the last part.


IH-WebDev-TA-Remote:
Good job on the lab! I see you were stuck in the end. One issue I noticed is that in lines 66 and 67 you misspell math, it should be Math. That should call the receiveDamage method correctly.
But good job overall, keep it up!

PR #1886, opened 2020-07-16
[AMS - PT - June20] - Kenouly

Kenouly:


src/viking.js
```diff
@@ -1,11 +1,78 @@
...
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health, strength)
+        this.name = name
+    }
+    receiveDamage(damage) {
```
guido-an, line 21: You could also reuse the receiveDamage function from Soldier using _super_: 
` receiveDamage (damage) {`
  `  super.receiveDamage(damage)`
   ` if (this.health > 0) {`
  `    return ${this.name} has received ${damage} points of damage`
 `   } else {`
 `     return ${this.name} has died in act of combat`
 `   }`
`  }`

src/viking.js
```diff
@@ -1,11 +1,78 @@
...
+            this.saxonArmy.splice(randomSaxon, 1);
+        }
+        return saxonDamage
+    }
+    saxonAttack() {
+        const randomSaxon = this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)];
+        const randomViking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
+        const vikingDamage = randomViking.receiveDamage(randomSaxon.attack());
+
+        if (randomViking.health < 0) {
```
guido-an, line 73: I could see there is only this test failing "War Armies Attack saxonAttack() method should remove dead vikings from the army".

If you just add the the '=' to the condition you should pass it: ` if (randomViking.health <= 0) {`

PR #1888, opened 2020-07-17
[PTAMS062020] Mahmut Sarcan

mhmtdl:


src/viking.js
```diff
@@ -1,11 +1,100 @@
...
+        return 'Odin Owns You All!'
+    }
+  
+
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    constructor() {
```
guido-an, line 44: Saxon should inherit from Soldier. Remove this constructor and you will pass 32 more tests

guido-an:
Well done Mahmut!

src/viking.js
```diff
@@ -1,11 +1,100 @@
...
+class Viking extends Soldier {
+    constructor(name,health,strength){
+       super(health,strength)
+       this.name = name
+
+    }
+    attack() {
+        return this.strength
+    }
+    receiveDamage(damage) {
```
guido-an, line 26: You could also reuse the receiveDamage function from Soldier using _super_: 
` receiveDamage (damage) {`
  `  super.receiveDamage(damage)`
   ` if (this.health > 0) {`
  `    return ${this.name} has received ${damage} points of damage`
 `   } else {`
 `     return ${this.name} has died in act of combat`
 `   }`
`  }`

guido-an:
Well done Mahmut!

PR #1930, opened 2020-08-19
WDTF-Berlin-Yoseob Shin - done except last bonus

shinyo627:


shinyo627:
FINIShed through all the bonuses

PR #1964, opened 2020-09-01
[PAR-WDFT-248] Laura D & Jorell

JorellLangmar:


N10T:
Hey Lisa & Marina,

**Nice commits**

**iteration 1: Soldier**
-
- 👍 

**iteration 2: Viking**
-
- try to keep the if(condition) on the same line with his action if you don't use curly brackets
`if(condition) action on the same line`

**iteration 3: Saxon**
-
- 👍 

**BONUS - Iteration 4: War 🥇 **
-
- 👍 
- `iSaxon`, try to be more explicit 
- no need the line 86 as condition, it's an else

** 🥇  SUPER BONUS 🥇  **
-
- No SuPeR BoNuS

Good job! 💯

PR #1978, opened 2020-09-21
 [REM-2009]zuzana

ZuzanaAK:
Hi Jorge, I have decided to leave this without working on the bonuses for now (because this topic is much clearer than the pokemon JS), and instead I will get through the materials for tomorrow and check the pokemon JS lab today. Thanks!

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,80 @@
 // Soldier
-class Soldier {}
+class Soldier {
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
No problem, definitely the mandatory part done means you know how classes work which is the main reason for this Lab. So Good job! Keep it up! :)

PR #1979, opened 2020-09-21
[REM 2009] Ross Homaidan

ibirossi:
I felt good about this Lab, it made more sense... up until the bonus.  I think I should be calling the previous functions but I cant work out how to do it.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,79 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength){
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Good job! the last part is definitely hard so don't worry too much, still give it a try when you have a chance :) Also, remember to make sure all your code is well indented for readability. everytime you open curly brackets for multiple lines, the elements should be indented one tab to the side.

Other than that, good work! keep it up!

PR #1980, opened 2020-09-21
[REM 2009]Marta Obregon

MartaObregon:


IH-WebDev-TA-Remote:
Good job on the mandatory parts, I would only suggest that you get used to the syntax of creating the methods outside the constructor, as it is more commonly used.

Your approach of the first part of the bonus was good! you only needed to do this.vikingArmy and this.saxonArmy for the properties, same as you did in previous iterations.

PR #1982, opened 2020-09-21
[REM-2009] Sandra Mendez

scmendez:


IH-WebDev-TA-Remote:
Very nice solution!! all iterations done, even the bonus which was hard. Code easy to read and very clean. Keep up the great work! :)

PR #1983, opened 2020-09-21
[REM-2009] Santiago Andrés Ardiles Páez

SantAndresP:
I don't know if I should declare functions inside **_constructor()_** or not.


IH-WebDev-TA-Remote:
Great work, all tests passing and code very clean and easy to read. Keep it up!
Doing the functions inside the constructor is perfectly fine but it is more common to write them outside of it without the this keyword, mostly for readability.

PR #1984, opened 2020-09-21
[REM-2009] Sumiya Ushiro

sumi0820:
Since I have hardly seen classes, it was a little bit hard to grasp the concept and logic. Although I managed to complete the LAB for today, I feel I need to practice more to fully understand and use it.

The other challenge for today was recursion. I understood the logic and why it happened, but again, I need to practice more to use it with more flexibility.

As for the LAB for today, I came up with one idea which I left with commented out lines(115~143). What I was thinking is that depending on the arguments passed to the function, each receiveDamage is invoked. Those random index lines can be shared, but the receiveDamage needs to be fired depending on which army attacks. The codes I left is just a rough idea, thus it might not be working though.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,143 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+  receiveDamage(damage) {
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Very nice solution! all the tests passing and code clean and easy to read.
Even the extra bonus done and is looking good, it should work :) Kudos
Keep up the great work!

PR #1985, opened 2020-09-21
[ REM 2009 ] Jordane Ruiz

jordaneruiz:
…rules of the exercice

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,103 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health; 
+        this.strength = strength; 
+
+        this.attack = function(){
+            return this.strength;
+        }
+        this.receiveDamage = function(damage){
+            this.health = this.health - damage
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Actually I think you understood everything good! I see most of the issues you had were syntax errors or forgetting to add simple lines of code :)

- in both methods for attack you are trying to target both a randomSaxon and a randomViking, so you need to get both random numbers and elements in each method. You can copy/paste the same lines, just change the name of the randomIndex variable to be different for each one.
- it is receiveDamage, not recieveDamage.
- inside the condition to remove the dead vikings/saxons target the health property of the randomSaxon/randomViking, not of the result, since result is just a string :)
- if their health is 0 it also means they are dead, so the comparison should be "<="

These are common issues during this lab, I think your approach, code logic and code cleanness is excelent! you just need to check the Jasmine errors and try to analyze why they are happening :)

But overall, getting to this point is good! you were very very close to finishing the bonus, keep it up!

PR #1986, opened 2020-09-21
[REM 2009] Aleksandra Tanasijenko

alekstana:
Yohooo

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,109 @@
 // Soldier
-class Soldier {}
+class Soldier {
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
Looping over the array to remove the dead vikings/saxons is clever, however to simplify the code, you already know the index of the saxon/viking that was attacked (index1, index2) so you could check only their health and remove it if needed.

Still overall, Awesome job! all tests passing :) Also code is easy to read and understand. Keep working this hard!

PR #1987, opened 2020-09-21
[REM - 2009] Vykintas Petrikas

VykiPetr:
Sorry its not finished, started very late and right now I am very very  tired. On removing the dead saxons/vikings it jasmine turns green only in this instance, when the health is above 0 if it 0 it says cant get health of undefined, not sure how to combat that. Ill finish it up! Sorry that its halfway

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,84 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health
+        this.strength = strength
+    }
+    attack(){
+        return this.strength
+    }
+    receiveDamage(dmg){
+        this.health -= dmg
+    }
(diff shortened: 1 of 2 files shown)


IH-WebDev-TA-Remote:
There are just a couple of things breaking there, here are some tips:

- the receiveDamage is the method that will subtract the health from the viking/saxon but this is executed after the conditional that checks if they died. But also you need to return it's result at the end! tricky...
- the rng you are trying here will work for the tests, but the result will always be the index 0 right? Math.random is a random number between 0 and 0.99999...
- what should happend if a saxon/viking health is exactly 0?

I think you did very good though, you could easily move onto the next topics/labs but still if you want to try to solve those, go for it :) Keep working hard!

PR #1988, opened 2020-09-22
[REM 2009] lovro hodak

LovroHodak:


IH-WebDev-TA-Remote:
Good solution! Almost all tests passing!

The issue that was breaking the test "should remove dead vikings from the army" was in line 21 and 39. it should be this.health. Otherwise it will point not to the current health but to the one when it was created.

And for the last tests, you just switched the strings to return :) 

But still, overall good approach, keep it up!

PR #2044, opened 2020-10-16
[FTAMS1020202] Nina van den Braak

ninsajala:
Iteration 1,2,3 done, now over to War


src/viking.js
```diff
@@ -1,11 +1,125 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health,
+            this.strength = strength
```
wax-bear, line 5: The indentation seems off here.

src/viking.js
```diff
@@ -1,11 +1,125 @@
...
+    }
+
+    attack() {
+        return this.strength
+
```
wax-bear, line 10: Please be mindful about the use of empty lines. Does this empty line improve the readability of the code?

src/viking.js
```diff
@@ -1,11 +1,125 @@
...
+    addViking(Viking) {
+        this.vikingArmy.push(Viking)
+    }
+
+    addSaxon(Saxon) {
+        this.saxonArmy.push(Saxon)
+    }
+
+    vikingAttack() {
+        let saxon = randomIndex(this.saxonArmy)
```
wax-bear, line 73: Will the value of the variable saxon be reasigned? If not, would, perhaps, a const be more apt here?

src/viking.js
```diff
@@ -1,11 +1,125 @@
...
+        let viking = randomIndex(this.vikingArmy)
```
wax-bear, line 74: Nice you are making use of a function here. Could the name, perhaps, be improved? What is this function ultimately doing? Are we getting something particular? If so, could that be processed in the function name?

src/viking.js
```diff
@@ -1,11 +1,125 @@
...
+            return 'Saxons have fought for their lives and survived another day...'
+        }
+        if (this.vikingArmy.length > 0 && this.saxonArmy.length > 0) {
+            return 'Vikings and Saxons are still in the thick of battle.'
+
+        }
+    }
+}
+
+function randomIndex(array) {
```
wax-bear, line 109: As discussed above, the naming could be altered. For example "getRandomWarrior" could do. Wouldn't it be better to make that function a method of the War class?

src/viking.js
```diff
@@ -1,11 +1,125 @@
...
+    return array[Math.floor(Math.random() * array.length)]
+}
+
+function cleanDeads(array) {
```
wax-bear, line 113: Same question regarding function/class method applies to cleanDeads.

src/viking.js
```diff
@@ -1,11 +1,125 @@
...
+    return array.filter(item => item.health > 0)
```
wax-bear: Imagine the army which is generated is big, contains many items. Filter, in that case, would be more costly than simply splicing the particular warrior based on the index which is used before to select one from the viking/saxon warriors array.

https://stackoverflow.com/a/44435418/3563737

A workaround would be to generate the index outside of the previous function and to store it in a variable. Then, the index can be submitted as a second argument to that function, and then later also be used here to splice the particular element.

src/viking.js
ninsajala: Check

src/viking.js
ninsajala: Will keep this in mind more, I tend to forget when I'm trying to solve the iterations. (Any extensions that can help me with it?)

src/viking.js
ninsajala: Yes, it's reassigned when the saxon is receiving damage

src/viking.js
ninsajala: Check

wax-bear (commented):
Good stuff Nina!
Please check out my comments for potential improvements.
Please reach out if you have any questions.

src/viking.js
wax-bear: I am not aware of such an extension unfortunately.

PR #2045, opened 2020-10-16
[FTAMS102020] Rich Alossi

alossix:
Required iterations complete, will work on bonus iterations

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -51,17 +51,48 @@ class War {
   addViking(Viking) {
     this.vikingArmy.push(Viking);
   }
+
   addSaxon(Saxon) {
     this.saxonArmy.push(Saxon);
   }
   vikingAttack() {
     // choose random sax
-    let vikingAttacker =
-      vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
+    let vikingAttacker = this.vikingArmy[
+      Math.floor(Math.random() * this.vikingArmy.length)
+    ];
     let saxonDefender = this.saxonArmy[
       Math.floor(Math.random() * this.saxonArmy.length)
     ];
+    let damage = saxonDefender.receiveDamage(vikingAttacker.strength);
+    if (saxonDefender.health <= 0) {
+      let i = this.saxonArmy.indexOf(saxonDefender);
+      this.saxonArmy.splice(i, 1);
+    }
+    return damage;
   }
-  //   saxonAttack()
-  //   showStatus()
+
+  saxonAttack() {
+    let saxonAttacker = this.saxonArmy[
+      Math.floor(Math.random() * this.saxonArmy.length)
+    ];
+    let vikingDefender = this.vikingArmy[
+      Math.floor(Math.random() * this.vikingArmy.length)
+    ];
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+  battleCry() {
+    return `Odin Owns You All!`;
+  }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  constructor(health, strength) {
+    super(health, strength);
```
wax-bear, line 36: Is this constructor function and declaration of the class properties necesary when only declaring those properties which are already set on the object parent, the extended class Soldier?

src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+    }
+    return `A Saxon has died in combat`;
+  }
+}
 
 // War
-class War {}
+class War {
+  vikingArmy = [];
+  saxonArmy = [];
```
wax-bear, line 50: These properties belong to the War class, correct? In that case, a constructor function would be useful here.

src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+  addViking(Viking) {
+    this.vikingArmy.push(Viking);
+  }
+
+  addSaxon(Saxon) {
+    this.saxonArmy.push(Saxon);
+  }
+  vikingAttack() {
```
wax-bear, line 58: On line 54 an empty line is used. Perhaps it would be more consistent to also use it after line 57.

src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+    // choose random sax
+    let vikingAttacker = this.vikingArmy[
```
wax-bear, line 60: The naming of the variables is expressive, which is a good practice!
Will the values of these (the current and following) variables be reassigned at some point? If not, would perhaps the use of const be more apt?

src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+      Math.floor(Math.random() * this.vikingArmy.length)
+    ];
+    let saxonDefender = this.saxonArmy[
+      Math.floor(Math.random() * this.saxonArmy.length)
+    ];
+    let damage = saxonDefender.receiveDamage(vikingAttacker.strength);
+    if (saxonDefender.health <= 0) {
+      let i = this.saxonArmy.indexOf(saxonDefender);
```
wax-bear, line 68: There appears to be a bit of double work going on here. On line 64, the index is generated by using the Math methods. Could, perhaps, the value of this generation be stored in a variable? This variable could be used on line 64 and in the splice method.

src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+  saxonAttack() {
+    let saxonAttacker = this.saxonArmy[
+      Math.floor(Math.random() * this.saxonArmy.length)
+    ];
+    let vikingDefender = this.vikingArmy[
+      Math.floor(Math.random() * this.vikingArmy.length)
+    ];
+    let damage = vikingDefender.receiveDamage(saxonAttacker.strength);
+    if (vikingDefender.health <= 0) {
+      let i = this.vikingArmy.indexOf(vikingDefender);
```
wax-bear, line 83: Please see my comment on line 68.

PR #2047, opened 2020-10-16
[FTAMS102020] Irene Asensio

ihatepineapple:


src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+    
+    addSaxon(Saxon) {
+        this.saxonArmy.push(Saxon)
+    }
+
+    vikingAttack() {
+        let saxon = this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)]
+        let viking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)]
+
+        saxon = saxon.receiveDamage(viking.strength)
```
wax-bear, line 69: We are know working in an object oriented way. Is this reassigning of the value necessary?

src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+        
+        this.saxonArmy = this.saxonArmy.filter(item => item.health > 0)
+        //this.saxonArmy = this.saxonArmy.filter(item => item.health > 0)
```
wax-bear, line 72: Is this comment still of any use?

src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+        return saxon
+        
+    }
+
+    saxonAttack() {
+        let viking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)]
+        let saxon = this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)]
+
+        viking = viking.receiveDamage(saxon.strength)
```
wax-bear, line 81: Same question here regarding the reassigning.

wax-bear (commented):
The test cases all pass. Great stuff!
I placed a few comments regarding the way you coded it.

PR #2068, opened 2020-10-17
[FTAMS2020]issaGruber

GruberI:


src/viking.js
```diff
@@ -1,11 +1,80 @@
...
+    addViking(Viking){
+        this.vikingArmy.push(Viking);
+    }
+
+    addSaxon(Saxon){
+        this.saxonArmy.push(Saxon);
+    }
+
+
+    //This is where I got stuck :(
```
wax-bear, line 68: This is a challenging part of the assignment!

Let me help you in pseudocode. What we would like to do is the following:
- Select a (random) viking from the vikingArmy
- Select a (random) Saxon from the saxonArmy
- Utilize the viking's receiveDamage method and the Saxon's attack method to subtract health points from the viking's health
- If the viking has no health points left, it should be removed from the vikingArmy array

PR #2070, opened 2020-10-17
[AMS 102020] Abbie Summers

ANSummers:
Completed first iteration. passed all jasmine criteria.


wax-bear (commented):
Good stuff Abbie!
For further improvement, I would uncomment the tests in the test file. Then, when opening SpecRunner.html, the test output can be used for iterating towards better outcomes.
If you have questions please reach out to me!

PR #2074, opened 2020-10-21
[FT-AMS-102020] Emad Ghnim

Emad-doz:


src/viking.js
```diff
@@ -1,11 +1,91 @@
...
+    constructor (health ,strength){
+        this.health= health,
+        this.strength= strength
+    }
+
+    attack(){
+        return this.strength;
+    }
+    receiveDamage(damage){
+        this.health = this.health - damage;
```
reMRKableDev, line 12: **A tip:** You could use an assignment operator here to do the same thing. This way you won't need to write the ```this.health``` twice.  Check suggestion below
```suggestion
        this.health -= damage;
```

src/viking.js
```diff
@@ -1,11 +1,91 @@
...
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength){
+        super(health, strength); 
+        this.name = name
+        }
+    receiveDamage(damage){
+        this.health = this.health - damage;
```
reMRKableDev, line 23: Same comment here as above 😄

src/viking.js
```diff
@@ -1,11 +1,91 @@
...
+        return 'Odin Owns You All!';
+    }
+}
+
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    receiveDamage(damage){
+        this.health = this.health - damage;
```
reMRKableDev, line 35: Here as well

src/viking.js
```diff
@@ -1,11 +1,91 @@
...
+        let randomSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+        let randomViking = Math.floor(Math.random() * this.vikingArmy.length);
+
+        let attackedSaxon = this.saxonArmy[randomSaxon].receiveDamage(
+          this.vikingArmy[randomViking].strength
+        );
+        if (this.saxonArmy[randomSaxon].health <= 0 ){
+         this.saxonArmy.splice(randomSaxon, 1)
+        } 
+        console.log(this.saxonArmy);
```
reMRKableDev, line 65: When submitting finished work, make it a habit to remove any ```console.log()``` in your work. 

**Why?**
For 2 reasons:
1️⃣ The ```console.log``` doesn't add any value to the code at this point since you already have an idea of what ```this.saxonArmy``` is.
2️⃣ The ```console.log``` won't add any value if this piece of code happens to go to a production environment. When you later work in a company, there will most likely already be tests that are looking at what this value is. In that regard, you wouldn't need to have the console.log

reMRKableDev (commented):
Nice work @Emad-doz 🚀 

Everything is in order, as far as the code goes. I left some tips that you can check out 😄

src/viking.js
Emad-doz: thanks for your feedback , will work on that and keep it in mind always

PR #2075, opened 2020-10-22
[FTAMS102020]Mahalakshmi

mahavinay:
Completed till Iteration Vikingattack method

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,105 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health
+        this.strength = strength
+    };
 
+    attack(){
+        return this.strength
+    };
+
+    receiveDamage(damage){
+        this.health = this.health - damage
+    };
+}
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name,health, strength){
+        super(health, strength);
+        this.name=name;
+    }
+
+
+    receiveDamage(damage) {
+        this.health = this.health - damage
+        if(this.health > 0){
+            return `${this.name} has received ${damage} points of damage`
+        } else { 
+            return `${this.name} has died in act of combat`
+        }
+
+    }
+
+    battleCry(){
+        return "Odin Owns You All!"
+    }
+
+
+
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier{
+    constructor(health, strength){
+        super(health, strength);
+    }
 
-// War
-class War {}
+    receiveDamage(damage) {
+        this.health = this.health - damage
+
+        if(this.health > 0){
+            
+            return `A Saxon has received ${damage} points of damage`
+        } else { 
+            return `A Saxon has died in combat`
+        }
+}
+}
+// Bonus War
+ class War {
+    constructor(){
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+        this.health = health
+        this.strength = strength
+    };
 
+    attack(){
+        return this.strength
+    };
+
+    receiveDamage(damage){
+        this.health = this.health - damage
```
reMRKableDev, line 13: **A tip:** You could use an assignment operator here to do the same thing. This way you won't need to write the ```this.health``` twice.  Check suggestion below
```suggestion
        this.health -= damage;
```

src/viking.js
```diff
@@ -1,11 +1,105 @@
...
-class Viking {}
+class Viking extends Soldier {
+    constructor(name,health, strength){
+        super(health, strength);
+        this.name=name;
+    }
+
+
+    receiveDamage(damage) {
+        this.health = this.health - damage
```
reMRKableDev, line 25: Same comment as above

src/viking.js
```diff
@@ -1,11 +1,105 @@
...
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier{
+    constructor(health, strength){
+        super(health, strength);
+    }
+
+    receiveDamage(damage) {
+        this.health = this.health - damage
```
reMRKableDev, line 49: same comment as above

src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+
+        if(this.health > 0){
+            
+            return `A Saxon has received ${damage} points of damage`
+        } else { 
+            return `A Saxon has died in combat`
+        }
```
reMRKableDev, line 56: **ES6 Tip**: Since this conditional statement only executes one thing in either scope, you can use a ternary operator here. See suggestion below
```suggestion
        return this.health > 0 ? `A Saxon has received ${damage} point of damage` : `A Saxon has died in combat`;
```

It saves you a few lines of code also 👍🏾

src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+}
+}
+// Bonus War
+ class War {
+    constructor(){
+        this.vikingArmy = [];
+        this.saxonArmy = [];
+       // super(health,strength)
+        //super.receiveDamage();
+        //this.strength=this.strength
```
reMRKableDev, line 66: You can remove these commented lines as you are not using them anywhere in your code

src/viking.js
```diff
@@ -1,11 +1,105 @@
 // Soldier
```
reMRKableDev, line 1: **A general tip**: The formatting of your document can be made easier if you install & use the extension "Prettier" for VsCode. Search for "Prettier" in your extension marketplace on vscode or find it here --> https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode 

You can also find how you install and use Prettier here --> https://www.youtube.com/watch?v=h3PJjP0nE98&ab_channel=JamesQQuick

reMRKableDev (commented):
Nice work @mahavinay 🚀 

I have left some tips on the work you have submitted so far. Check them out!
Also, try and finish the last parts of this lab 😄

PR #2076, opened 2020-10-22
[FTAMS102020] Ahmet Demirlicakmak

codebrokerad:
All complete

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,83 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+    }
+    attack() {return this.strength}
+    receiveDamage(damage) {this.health=this.health-damage; }
+    
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health, strength);
+        this.name = name; 
+    }
+
+    receiveDamage(damage){
+        this.health = this.health-damage;
+    if (this.health > 0){
+        return `${this.name} has received ${damage} points of damage`;
+    } else {
+        return `${this.name} has died in act of combat`;
+    }
+    }
+    battleCry(){
+        return "Odin Owns You All!";
+    }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    receiveDamage(damage){
+        this.health = this.health-damage;
+    if (this.health > 0){
+        return `A Saxon has received ${damage} points of damage`;
+    } else {
+        return "A Saxon has died in combat";
+    }
+}
+}
+    
+
 
 // War
-class War {}
+class War {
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,83 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+    }
+    attack() {return this.strength}
+    receiveDamage(damage) {this.health=this.health-damage; }
```
reMRKableDev, line 8: **A tip:** You could use an assignment operator here to do the same thing. This way you won't need to write the ```this.health``` twice.  Check suggestion below
```suggestion
    receiveDamage(damage) {this.health -= damage;}
```

src/viking.js
```diff
@@ -1,11 +1,83 @@
...
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health, strength);
+        this.name = name; 
+    }
+
+    receiveDamage(damage){
+        this.health = this.health-damage;
```
reMRKableDev, line 20: Same comment as above

src/viking.js
```diff
@@ -1,11 +1,83 @@
...
+    if (this.health > 0){
+        return `${this.name} has received ${damage} points of damage`;
+    } else {
+        return `${this.name} has died in act of combat`;
+    }
```
reMRKableDev, line 25: **ES6 Tip**: Since this conditional statement only executes one thing in either scope, you can use a ternary operator here. See suggestion below
```suggestion
    return this.health > 0 ? `${this.name} has received ${damage} points of damage` : `${this.name} has died in act of combat`;
```

It saves you a few lines of code also 👍🏾

src/viking.js
```diff
@@ -1,11 +1,83 @@
...
+        return "Odin Owns You All!";
+    }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    receiveDamage(damage){
+        this.health = this.health-damage;
+    if (this.health > 0){
```
reMRKableDev, line 36: Same comment here as the ternary operator above

src/viking.js
```diff
@@ -1,11 +1,83 @@
 // Soldier
```
reMRKableDev, line 1: **A general tip**: The formatting of your document can be made easier if you install & use the extension "Prettier" for VsCode. Search for "Prettier" in your extension marketplace on vscode or find it here --> https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode 

You can also find how you install and use Prettier here --> https://www.youtube.com/watch?v=h3PJjP0nE98&ab_channel=JamesQQuick

reMRKableDev (commented):
Nice work @codebrokerad 🚀 

Your solutions are all nicely done. I've left some tips that you can check out 😄

PR #2084, opened 2020-10-23
[WDFT-AMS-102020]-Alex Kantsadis

teof95:
it is not complete. missing few things. I need to update

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -21,12 +21,21 @@ class Soldier {
     }
 }
 // Viking
-class Viking {
-    constructor(name) {
+
+class Viking extends Soldier {
+    constructor(health, strength, name) {
+        super(health, strength)
         this.name = name;
     }
 }
-attack()
+
+
+// class Viking {
+//     constructor(name) {
+//         this.name = name;
+//     }
+// }
+// attack()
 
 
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,65 @@
...
+    }
+
+    receiveDamage(damage) {
+        this.health -= damage;
+        if (this.health > 0) {
+            return this.name + " has received " + damage + " points of damage";
+        } else {
+            return this.name + " has died in act of combat";
+        }
+    }
```
LloydC, line 38: Nice, you can also write line 33-37 using a ternary operator
```
return this.health > 0 ? this.name + " has received " + damage + " points of damage" : this.name + " has died in act of combat";

File: src/viking.js [hunk 2]
```diff
@@ -1,11 +1,65 @@
...
+        super(health, strength)
+    }
+    receiveDamage(damage) {
+        this.health -= damage;
+        if (this.health > 0) {
+            return "A Saxon has received " + damage + " points of damage";
+        } else {
+            return "A Saxon has died in act of combat";
+        }
+    }
```
LloydC, line 58: Same comment as above

PR #2092, opened 2020-10-23
[FTRMT102020] - Marcel Niebylski

mnblski:
All done

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,103 @@
+
+
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+
+    attack() {
+        return this.strength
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,103 @@
...
+        
+        let attack = randomViking.receiveDamage(randomSaxon.strength);
+
+        this.vikingArmy = this.vikingArmy.filter(viking => viking.health > 0);
+
+        return attack
+        }
+    
+    showStatus() {
+        if (this.saxonArmy == 0) {
```
filipe-freire, line 93: Although in this lab it is not too relevant, the strictly equals (===) should be used as a good practice, in order to prevent unwanted coercions!

filipe-freire (approved):
Fantastic job Marcel, nice work! 🎉

PR #2117, opened 2020-10-24
[WDFTRM202010](24/10/20)Hugo

h-u-g-o-1989:


src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+  receiveDamage(damage) {
+    this.health = this.health - damage;
```
filipe-freire, line 11: could also use the shorthand `this.health -= damage` to achieve the same purpose!

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+      Math.floor(Math.random() * this.saxonArmy.length)
+    ];
+    const randomVik = this.vikingArmy[
+      Math.floor(Math.random() * this.vikingArmy.length)
+    ];
+    const attackResult = randomSax.receiveDamage(randomVik.strength);
+    if (randomSax.health <= 0) {
+      // find random saxon in saxon army array, going to use the index of method, then we will need to splice as that allows us to delete a specific index in the array and mutates the orignal array rather than returning a new one
+      const randomSaxonIndex = this.saxonArmy.indexOf(randomSax);
+      this.saxonArmy.splice(randomSaxonIndex, 1);
```
filipe-freire, line 71: A completely good solution to this problem! You could also use the filter method on the army array to check if there's any dead warrior!

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+      Math.floor(Math.random() * this.vikingArmy.length)
+    ];
+    const attackResult = randomVik.receiveDamage(randomSax.strength);
+    if (randomVik.health <= 0) {
+      const randomVikingIndex = this.vikingArmy.indexOf(randomVik);
+      this.vikingArmy.splice(randomVikingIndex, 1);
+    }
+    return attackResult;
+  }
+  // again probably a better way of checking than using if statements here, sorry if this is not the logic im meant to be using.
```
filipe-freire, line 90: if statements are just fine here, although you could also use `if else` statements since you only want one of them happening at a time

filipe-freire (approved):
Awesome job solving the lab! Congratulations! 🎉

PR #2121, opened 2020-10-24
[DWFTRM - 20201024] - Gosia

GosiaMas:
Super bonus - showStatus function works when it is a snippet, but not passes Jasmin.

I struggle with attack functions

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,97 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
 
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+  }
+}
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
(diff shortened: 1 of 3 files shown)


src/viking.js
```diff
@@ -1,11 +1,106 @@
...
+    //   Math.floor(Math.random() * saxonArmy.length + 1)
+    // ].receiveDamage(
+    //   this.vikingArmy[Math.floor(Math.random() * vikingArmy.length + 1)]
+    // );
+
+    // if ((randomSaxon.health = 0)) {
+    //   this.saxonArmy.splice(randomSaxon, 1);
+    // }
+
+    this.saxonArmy.filter((eachSaxon) => eachSaxon.health > 0);
```
filipe-freire, line 88: Awesome use of the filter method! But since filter always returns an array, you should assign the "this.saxonArmy" to the filtered array in order for your solution to work!

filipe-freire (commented):
Excellent work Gosia! Made a comment on the usage of the filter method, but other than that, excellent work!

src/viking.js
GosiaMas: Hi Filipe, I did this, but still 8 failures...

this.saxonArmy = this.saxonArmy.filter((eachSaxon) => eachSaxon.health > 0);

PR #2124, opened 2020-10-24
w

emmaryal:


src/viking.js
```diff
@@ -1,11 +1,110 @@
...
+  addViking(viking) {
+    this.vikingArmy.push(viking);
+  }
+
+  addSaxon(saxon) {
+    this.saxonArmy.push(saxon);
+  }
+
+  vikingAttack() {
+    var randomViking = this.vikingArmy[
```
MartaJank, line 65: You should stop using 'var' and start using "let" or "const" depending on which one is more appropriate for each occasion.

PR #2135, opened 2020-10-25
[WDFTRM-20201025] Marc Gotzens

Mgot12602:
Finished but still Bonus

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,63 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+}
+    attack() {
+        return this.strength;
+}
+    receiveDamage(damage) {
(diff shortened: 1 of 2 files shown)


src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+    constructor(name, health, strength) {
+        super(health, strength);
+        this.name = name;
+        }
+        receiveDamage(damage){
+            super.receiveDamage(damage);
+            if (this.health > 0) {
+                return `${this.name} has received ${damage} points of damage`;
+            } return `${ this.name} has died in act of combat`;
+        }
```
filipe-freire, line 27: In this code block, you can erase the `super.receiveDamage(damage);`. The reveiceDamage method should subtract the passed damage from the health property of the Viking object!

filipe-freire (commented):
Overall, very nice work Marc! 😀

PR #2140, opened 2020-10-25
BCN WBOCT2020 Sergio Cabrera

capineti:


src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+    return this.strength;
+  }
+
+  receiveDamage(damage) {
+    this.health = this.health - damage;
+  }
+}
 // Viking
-class Viking {}
+class Viking extends Soldier {
```
MartaJank, line 17: I've cleaned up a little your code here and it works like that. The main problem vas to use the Arg particle on the end of each argument, if you delete that all over your code it automatically works well.

    super(health, strength);
    this.name = name;
  }
  receiveDamage(damage) {
    this.health = this.health - damage;
    if (this.health > 0) {
    } else {
    }
  }
  battleCry() {
    return `Odin Owns You All!`;
  }
}

src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+  constructor(nameArg, healthArg, strengthArg) {
+    super(healthArg, strengthArg);
+    this.name = nameArg;
+  }
+  receiveDamage(damage) {
+    this.health = this - damage;
```
MartaJank, line 23: Also, here you just forgot to put "this.health" after the equal sign, you just wrote "this".

PR #2143, opened 2020-10-25
[BCN - FT - 102020] _ Sergi Llorente

sergillorente:


src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+            return `A Saxon has died in combat`
+        }
+        }
+}
 
 // War
-class War {}
+class War {
+    constructor() {
+        vikingArmy: [],
```
MartaJank, line 51: The comma you have here should be a semicolon (;), it is throwing an error otherwise and it does not allow you to pass any Jasmine tests.

src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+        saxonArmy: []
+    }
+    addViking(viking) {
+        this.vikingArmy = 1;
+    } 
+    addSaxon(saxon) {
+        this.saxonArmy = 1;
+    }
+    vikingAttack()
```
MartaJank, line 60: If you don't see your tests passing still, just comment this three functions that are not yet written in order to see the tests passing (if you want to check on your code).

PR #2147, opened 2020-10-26
Don't pass the test but working for me

mirko1075:


mirko1075:
Inside the vikingAttck and saxonAttack methods I didn't create the soldiers objs since I retrieve them from the army arrays.

PR #2151, opened 2020-10-26
WDFTRM - 20201026 -Querien

querien:


src/viking.js
```diff
@@ -1,11 +1,88 @@
...
 // War
-class War {}
+class War {
+  constructor() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+  }
+
+  addViking(Viking) {
+    this.vikingArmy = [Viking];
```
filipe-freire, line 58: The addViking method should **push** a Viking into the existing vikingArmy array! With this approach, what is happening is that you are setting the vikingArmy property to an array of a Viking, whenever the method is called!

filipe-freire (commented):
An overall very good job! All mandatory iterations completed perfectly 🥇

PR #2283, opened 2021-01-19
FT-BER-012021 - Hong Hai Nguyen - All done!

honghainguyen777:


honghainguyen777:
I forget to use the attack() method in the Soldier class to get the strength. In stead of the, I call the strength property directly. Sorryyyyyyyyu

PR #2357, opened 2021-03-26
[FTAMS032021] Andrea Sforza

arsforza:
All mandatory and bonus assignments completed.


reMRKableDev (commented):
Nice work @arsforza 🚀 

Good implementations and well used refactoring

PR #2367, opened 2021-03-26
[FTAMS032021] Jeroen, Andrea Ayoub

jeroenvanbart:


src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength =  strength;
+    }
+    attack() {
+        return this.strength;
+    }
+    receiveDamage(damage) {
+        let currentHealth = this.health - damage;
```
reMRKableDev, line 11: You don't need a new variable here. We need to reference `this.health`

```suggestion
        this.health = this.health - damage;
```

src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+    }
+}
+
+
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(health, strength, name){
```
reMRKableDev, line 19: `name` should be the first parameter in the constructor

```suggestion
    constructor(name,health, strength){
```

src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+    super (health, strength)
+        this.name= name;
+    }
+    receiveDamage(damage){
+        let currentHealthViking = this.health - damage;
+        if (currentHealthViking > 0){return `${this.name} has received ${this.damage} points of damage`}
+        else if (currentHealthViking <= 0){return `${this.name} has died in act of combat`}
```
reMRKableDev, line 26: Make use of `this.health` instead of using a new variable. Otherwise, the information on the output won't match

```suggestion
    this.health = this.health - damage;
    if (this.health > 0) {
      return `${this.name} has received ${damage} points of damage`;
    } else {
      return `${this.name} has died in act of combat`;
    }
```

src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+    }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    receiveDamage(damage){
+        let currentHealthSaxon = this.health - damage;
+        if (currentHealthSaxon > 0){return `A Saxon has received ${this.damage} points of damage`}
+        else if (currentHealthSaxon <= 0){return `A Saxon has died in act of combat`}
```
reMRKableDev, line 40: Same comment as above

reMRKableDev (commented):
Nice work @jeroenvanbart 🚀 

I've added some comments for you to make adjustments to your code. Check with the class we shared in class to see how War would work

PR #2502, opened 2021-06-07
WDFT-AMS-052021 <AnnaSun>

AnnaSun919:
solved the mandatory part


TA-Remote:
Awesome Anna!! Well done! 💯 You also tried the war bonus 👍 Keep the hard work! Could have tried to give a shot to the last function (showStatus!) is the only thing you are missing! 🥇

PR #2508, opened 2021-06-07
 [WDFT-RMT-MAY21] Conchita 

CodeChita:
Hi TA's! 

Hope you are having a lovely morning! I made the bonus in a strange way I think! Is there a way to make it more reusable? Love to get some feedback on that :)

Have a great day!

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -75,14 +75,14 @@ class War {
      return saxonResult
   }
   saxonAttack(){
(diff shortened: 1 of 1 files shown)


TA-Remote:
Very good job on the lab :) Feel free to check the solution later to see how you could have done the bonus but your solution is completely viable!

PR #2520, opened 2021-06-07
Solved Lab

ritamak:


TA-Remote:
Hey Rita! You did a really good job on the lab! Just for you to know, it wasn't necessary to call again the attack method in the Vikings class and the saxons class (it's written in the instructions). Very good job again, keep going!

PR #2524, opened 2021-06-07
solved

gigi-loranca:


TA-Remote:
Nice job on the lab. Be super careful when asked to write a specific message in a return. The tests are case sensitive and most of your mistakes are because you did not put an uppercase somewhere or a punctuation.

PR #2570, opened 2021-07-23
[WD-PT-Remote] Juan Carlos

juaneugenio:
I was not so sure, how can I invoke a "constructor" from an already created object. I guess, i can do that just invoking the "super method" from the alterated/new class?

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -20,7 +20,19 @@ class Viking extends Soldier {
     this.health = health;
     this.strength = strength;
   }
-  battleCry() {}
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
All good in here Juan! Great job once again! :tada: 
Regarding your doubt, you are correct. The `super()` expects as many arguments as the initial class needs in order to create its properties.

In this case, since `Viking` and `Saxon` both share the health and strength properties of the `Soldier`, we use the super() to pass those same values to its constructor :muscle:

PR #2574, opened 2021-07-27
[WD-PT-27/07/2021] Anna Velarde

annavelarde:


IH-WebDev-TA-Remote:
Nice job Anna! Another lab done! :muscle: 

Here's some feedback:
-> The super() method expects as many arguments as the initial class we are extending needs in order to create its properties! So in this lab, since Viking and Saxon both share the health and strength properties of the Soldier, we use the super() to pass those same values to its constructor. No need to declare them again. :wink:

PR #2586, opened 2021-08-07
Finished lab

samueloh99:


src/viking.js
```diff
@@ -1,14 +1,62 @@
...
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  constructor(health,strength){
+    super(health,strength)
+  }
+
+  receiveDamage(damage){
+    this.health -= damage
```
henriquemorikawa, line 46: super.receiveDamage(damage) 😉

src/viking.js
```diff
@@ -1,14 +1,62 @@
...
+    }else{
+      return "A Saxon has died in combat"
+    }
+  }
+}
 
 // War
-class War {}
+class War {
+
```
henriquemorikawa, line 58: bora terminar mano!!! Falta pouco

PR #2668, opened 2021-10-09
classes exercice

nguasch:


src/viking.js
```diff
@@ -1,14 +1,88 @@
...
+  receiveDamage(damage) {
+    this.health = this.health - damage
+  }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(Soldier);
```
tonyghiani, line 20: No need to Soldier to super() since it already get it from extending the class.

src/viking.js
```diff
@@ -1,14 +1,88 @@
...
+    }
+  }
+}
 
 // War
-class War {}
+
+
+class War {
+  vikingArmy = []
```
tonyghiani, line 59: These should be properties in the constructor

PR #2798, opened 2021-12-10
[WDPT BER 11.2021] Aljoscha Zynga

azynga:


src/viking.js
```diff
@@ -1,14 +1,107 @@
...
-class Viking {}
+class Viking extends Soldier {
+
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
```
numbing, line 28: if you see in soldier class we have the same method right? instead of writing line 28 we can use
this.receiveDamage(damage)

Not its only one line but you can imagine what if its bigger"normaly is bigger"

src/viking.js
```diff
@@ -1,14 +1,107 @@
...
+  }
+
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  
+  receiveDamage(damage) {
+    this.health -= damage;
```
numbing, line 46: here the same

PR #2812, opened 2022-01-14
[REMOTE FT 012022] Denise Sengül

denisesenguel:


luisjunco:
great job Denise!

on last iteration:
- it is possible to declare a function inside a method (just as you have it)
- in this particular case, code will be much easier to follow if you create `pickRandom()` as a separate method

Something like this:

```
  pickRandom(army) {
  }

  attack(whoIsAttacking) {

    //...
    //...

    const attacker = this.pickRandom(attackingArmy);
    const victim = this.pickRandom(victimArmy);

  }

```

PR #2813, opened 2022-01-14
Nico Tollenaar Lab Solved, including bonusses

NicoTollenaar:


luisjunco:
nice work Nico!


is there a way to avoid repeating [all this code](https://github.com/NicoTollenaar/lab-javascript-vikings/blob/0f0d7239cf3bf71267b6f15df5719917d36d92c9/src/viking.js#L69) to pick a random element from the army?


hint: 
```
pickRandomSoldier(army){
  //returns a random element from the army
}
```

NicoTollenaar:
Hi Luis, thanks for the feedback, agreed, much better and clearer with proposed extra method

PR #2814, opened 2022-01-14
[WDFT - Jan 22] Day 5 Lab 1 - Doan Tran

Dorrrn:
Iteration 1-3 done

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -89,16 +89,16 @@ class War {
 
     return vikingAttacked;
   }
-}
 
-// Show Status
-function showStatus() {
-  if (this.saxonArmy.length < 1 && this.vikingArmy.length > 0) {
-    return 'Vikings have won the war of the century!';
-  } else if (this.vikingArmy.length < 1 && this.saxonArmy.length > 0) {
-    return 'Saxons have fought for their lives and survived another day...';
-  } else if (this.saxonArmy.length === 1 && this.vikingArmy.length === 1) {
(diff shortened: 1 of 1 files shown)


Dorrrn:
wow, what a syntax challenge :D

src/viking.js
```diff
@@ -1,16 +1,108 @@
...
+// Show Status
+function showStatus() {
+  if (this.saxonArmy.length < 1 && this.vikingArmy.length > 0) {
+    return 'Vikings have won the war of the century!';
+  } else if (this.vikingArmy.length < 1 && this.saxonArmy.length > 0) {
+    return 'Saxons have fought for their lives and survived another day...';
+  } else if (this.saxonArmy.length === 1 && this.vikingArmy.length === 1) {
+    return 'Vikings and Saxons are still in the thick of battle.';
+  }
+}
```
TA-Remote, line 103: Your `showStatus `method should be inside War class ;)

TA-Remote (commented):
Great work Doan! 

In the last iteration, as Luis suggested today you can make code shorter and more readable by separatingthe generic logic. For example inside War class you can create   `pickRandom(army) {}` method . And call that method instead line 62, 65 (by passing the needed army) etc.

src/viking.js
Dorrrn: Ahhh !!!  that makes sense now :D

Dorrrn: As always, thanks a lot for reviewing !! :)

PR #2816, opened 2022-01-14
WTFT Remote Helena Saborit

Helsinky91:
still trying to figure out about how to express "alive"

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,14 +1,56 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+
+  attack() {
+    return this.strength
+  };
+
+  receiveDamage(damage) {
+    this.health -= damage
+  };
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (Viking === "alive"){
+      return (`${this.name} has receive ${this.damage}`);
+    } else {
+      return (`${this.name} has receive died in combat`)
+    }
+  };
+
+  battleCry(){
+    return ('Odin Owns You All!')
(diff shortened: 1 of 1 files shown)


TA-Remote:
Haha, I can relate! Ask next time in Slack, someone would have liked to help for sure!
I'll make you some comments after lunch and then everything will be clear to you.

src/viking.js
```diff
@@ -1,14 +1,56 @@
...
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (Viking === "alive"){
```
TA-Remote, line 26: Viking inherited "this.health" from Soldier. To see if the Viking is still alive you need to check the health of the Viking. If the number of health is lower or equal 0, the Viking died. If the health is bigger than 0, the Viking is still alive. In code:

```
  receiveDamage(damage) {
    this.health -= damage;
    if (this.health > 0){
      return (`${this.name} has received ${damage} points of damage`);
    } else {
      return (`${this.name} has died in act of combat`)
    }
  };

```

src/viking.js
```diff
@@ -1,14 +1,56 @@
...
+
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier { 
+  
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (Saxon === 'alive' ) {
+      return (`A Saxon has received ${this.damage} points of damage`)
```
TA-Remote, line 45: here and also in the Viking class it must be just "damage" because you give it as a parameter to the function (instead of creating it in the class in the constructor). So in code, this one you could have solved like this:
```
  receiveDamage(damage) {
    this.health -= damage;
    if (this.health > 0) {
      return (`A Saxon has received ${damage} points of damage`)
    } else {
      return (`A Saxon has died in combat`)
    }
  }
}
```

TA-Remote: If you have any questions about this, feel free to ask us!

PR #2842, opened 2022-01-16
WD Remote Jan2022 - Ana Pavão

pavaoana:
lab solved + a bit of the bonuses

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,16 +1,86 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,16 +1,86 @@
...
-
-
+class War {
+  constructor() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+  }
+  addViking(viking) {
+    this.vikingArmy = viking; // ??
+  }
```
TA-Remote, line 68: To add viking to an army, you would want to push it into your array

PR #2925, opened 2022-04-01
WDFTR MAR2022 Vikings

mafaldao21:
having some trouble making the receiveDamage() method work

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,14 +1,116 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+
+  attack() {
+    return this.strength
+
+  }
+
+  receiveDamage(damage) {
+    this.health =- damage;
+  }
+
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength){
+   super();
+   this.name = name;
+   this.health = health;
+   this.strength = strength;
+  }
+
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,14 +1,116 @@
...
+  }
+
+  attack() {
+    return this.strength
+
+  }
+
+  receiveDamage(damage) {
+    this.health =- damage;
+  }
```
TA-Remote, line 15: Little but important syntax mistake!  Should be ` this.health -= damage;`

To avoid this , do not hesitate and make it more obvious when you're not sure why it does not work. For example just right it like this and you won't be wrong:
 `this.health = this.health - damage;`

src/viking.js
```diff
@@ -1,14 +1,116 @@
...
+
+  receiveDamage(damage) {
+    this.health - damage;
+    if(this.health>0) {
+    console.log(`${this.name} has received ${damage} points of damage`);
+    } else {
+      console.log(`${this.name} has died in act of combat`)
+    }
+  }
```
TA-Remote, line 35: - On line 29, you are missing equals sign:` this.health -= damage;`
- Lines 31& 33, if you read the error in tests you can fix this ;) Task asks you to RETURN , and you are console.logging

`should return "NAME has received DAMAGE points of damage", if the Viking is still alive`

TA-Remote (commented):
Great work Mafalda! but many syntax errors, be attentive when you read task, and even more so read the errors you are getting, it points exactly to where the mistake is ;)

PR #2930, opened 2022-04-01
ALEJANDRO BLANCO WDFT MARCH 2022

abg1995:


src/viking.js
```diff
@@ -1,14 +1,85 @@
...
+
+}
 
 // War
-class War {}
+class War {
+  constructor() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+  }
```
TA-Remote, line 67: Good job, Alejandro!
For the addViking and addSaxon method:
to add a viking or a saxon, simply push the viking/saxon being passed as an argument to the vikingArmy/saxonArmy array, like this: 
```
  addViking(viking) {
    this.vikingArmy.push(viking);
  }
```

PR #2932, opened 2022-04-01
Solved lab

PJPalma:


src/viking.js
```diff
@@ -1,14 +1,88 @@
...
+      const randomViking = this.vikingArmy[Math.floor(Math.random(0, this.saxonArmy.length))]
+  
+      return randomSaxon.receiveDamage(randomViking.strength)
+  
+      if (randomSaxon.health <= 0) {
+        let indexViking = this.vikingArmy.indexOf(randomViking)
+        randomViking.splice(indexViking, indexViking)
+        return result
+      }
+    }
```
TA-Remote, line 82: If you have` return ` in the middle of your method, anything below it will not be read..

PR #2933, opened 2022-04-01
WDFT Remote Lab Viking - Ha Thu

HaThu6989:


src/viking.js
```diff
@@ -1,14 +1,75 @@
...
+  }
+
+  battleCry() {
+    return ("Odin Owns You All!")
+  }
+
+  receiveDamage(damage) {
+    this.health = this.health - damage
+    if (this.health > 0) {
+      return (`${this.name} has received ${this.health} points of damage`)
```
TA-Remote, line 31: Hi Thu, here the message is:
    `  return (`${this.name} has received ${damage} points of damage`)`
    damage instead of this.health

src/viking.js
```diff
@@ -1,14 +1,75 @@
...
+  }
+
+  battleCry() {
+    console.log("Odin Owns You All!")
+  }
+
+  receiveDamage(damage) {
+    this.health = this.health - damage
+    if (this.health > 0) {
+      return (`A Saxon has received ${this.health} points of damage`)
```
TA-Remote, line 56: same here:
```     return (`A Saxon has received ${damage} points of damage`)```

File: src/viking.js [hunk 3]
```diff
@@ -1,14 +1,75 @@
...
+    } else {
+      return ("A Saxon has died in act of combat")
+    }
```
TA-Remote, line 59: here it is just a small mistake in the sentence:
`return ("A Saxon has died in combat")`

PR #2938, opened 2022-04-02
WDFTRemoteMarch2022+Sepid.att

Sepidehatt:


src/viking.js
```diff
@@ -1,14 +1,116 @@
...
-class Saxon {}
+class Saxon extends Soldier {
+
+  constructor(health,strength){
+    super(health, strength);
+  }
+
+  receiveDamage(damage){
+    this.health -= damage ;
+    if(this.health !== 0) return `A Saxon has received ${damage} points of damage`;
```
TA-Remote, line 46: If you change this condition to this` if(this.health > 0)` , test is passing ;)

Sepidehatt: Thank you very much for your help..this part was very confusing to me

PR #2940, opened 2022-04-03
WDFT MAR 22 Harry

hmarchbank:
Changed my mind on the 'this' in final super bonus challenge

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -95,11 +95,11 @@ class War {
 
 
   attack(attackingArmy, defendingArmy){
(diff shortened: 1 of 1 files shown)


TA-Remote:
Awesome, Harry! You did almost the same as I did for the super bonus refactor thingy, but you can still compare it, if you want to:
[](https://github.com/Joanneseiler/lab-javascript-vikings/blob/master/src/viking.js)
Because we do not have a solution for this in the Student Portal.

PR #2943, opened 2022-04-03
WDFT March2022 Valentin Paul

Valentin-Paul:
Didn't get any further with the WAR Bonus... I will look it up in the solutions tomorrow

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,14 +1,71 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength){
+    this.health = health
+    this.strength = strength
+  }
+  attack(){
+    return this.strength;
+  }
+  receiveDamage (damage){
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,14 +1,71 @@
...
+  }
+}
 
 // War
-class War {}
+class War {
+constructor(vikingArmy, saxonArmy){
+  this.vikingArmy = vikingArmy
+  this.saxonArmy = saxonArmy
+}
```
TA-Remote, line 51: When we work with tests, make sure to read errors! Very helpful. For example, your first two errors for war class:
- should receive 0 arguments
- should assign an empty array to the vikingArmy property

these suggest following: 

```
constructor(){
  this.vikingArmy = []
  this.saxonArmy = []
}
```

PR #2947, opened 2022-04-04
vikings

marija-strah:


src/viking.js
```diff
@@ -1,14 +1,73 @@
...
+  addViking(Viking) {
+    this.vikingArmy.push(Viking);
+  }
+  addSaxon(Saxon) {
+    this.saxonArmy.push(Saxon);
+  }
+  vikingAttack() {
+    receiveDamage(callback);
+    this.saxonArmy(pop);
+  }
```
TA-Remote, line 67: In the attack method, first thing you need to do is to get random element from the army array.
So take viking army array and using Math.random get one element (one army) from it.

After that,  take saxon army array and get one random element.

Then we need to implement the following task:
`should make a Saxon receiveDamage() equal to the strength of a Viking`

This means:
- randomly chosen saxon army gets damaged / call receiveDamage
- damage is equal to result of calling viking attack method

PR #2955, opened 2022-04-15
[WDFT042022] Dino Marchiori

dinaco:


webLisbonTA (commented):
Great job on doing every mandatory iteration and the bonus!

src/viking.js
```diff
@@ -1,17 +1,89 @@
...
+    this.vikingArmy.push(newViking);
+  }
+  addSaxon(newSaxon) {
+    this.saxonArmy.push(newSaxon);
+  }
+  vikingAttack() {
+    const randomizerSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+    const randomizerViking = Math.floor(Math.random() * this.vikingArmy.length);
+    const randomSaxon = this.saxonArmy[randomizerSaxon];
+    const randomViking = this.vikingArmy[randomizerViking];
```
webLisbonTA, line 59: Here, if you wanted make a shorter solution, you could access the array instead of creating those variables:
```
const randomViking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)]
```

PR #2961, opened 2022-04-15
[FTLIS042022] Anna Egger

annagramcodes:
Wasn't sure how to refactor the two attack Methods, so they will be represented by one generic one.


webLisbonTA (commented):
Great job on doing every iteration including the bonus!

PR #2979, opened 2022-04-22
test2

zarahsara:


Enes-Ironhack:
You tried. But you are missing a lot by not asking questions... you have missing brackets everywhere => It results as a fail.
Please, ask questions if you have any doubt so that we can assist you better.
This was a better attempt than the last times, keep trying !

PR #2985, opened 2022-04-23
done, but don't know about return-statements

EvgeniiaRiaguzova:


Enes-Ironhack:
You have some typo errors here and there but it's a success overall ;) good job !

PR #3069, opened 2022-06-12
All main tasks are done

klwegner:
I finished the required tasks but couldn't figure out most of the last two (bonus) tasks.


syltiva:
Looks great! To avoid re-writing `this.health -= theDamage;` on lines 29 & 46, you can also use the 'super' keyword. 

From MDN: 
`super.functionOnParent([arguments]);`

PR #3086, opened 2022-06-17
Lara Rockenstein WD FT June 22

LMRBLN:


src/viking.js
```diff
@@ -1,14 +1,100 @@
...
+  saxonAttack() {
+    let randomIndexSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+    let randomIndexViking = Math.floor(Math.random() * this.vikingArmy.length);
+    
+    this.vikingArmy[randomIndexViking].receiveDamage(this.saxonArmy[randomIndexSaxon].strength);
+    if (this.vikingArmy[randomIndexViking].health<=0) {
+    this.vikingArmy.splice(randomIndexViking, 1);
+    return       //return this.vikingArmy[randomIndexViking].name + " has received " + damage + " points of damage";
+    ;
+    }
```
TA-Remote, line 81: To pass this last test, rememeber that you can save any result of calling a function in a variable, so for example, line 77 returns a string already -> you defined it in Saxon class:

```
  receiveDamage(damage) {
    this.health-=damage;
    if (this.health > 0) {
      return "A Saxon has received " + damage + " points of damage";
    }
    else {
      return "A Saxon has died in combat";
    }  
  }
```

so all you want to do is save this in a new variable 
```
const result = this.vikingArmy[randomIndexViking].receiveDamage(this.saxonArmy[randomIndexSaxon].strength);
```

and return this result at the end of the saxonAttack() method.
In other words calling saxonAttack() will return you a string

TA-Remote (commented):
Excellent

PR #3088, opened 2022-06-17
[RM FT June 22] Alex's Solved lab - vikings 

Virgule09:


TA-Remote:
Great!

src/viking.js
```diff
@@ -1,14 +1,91 @@
...
+  constructor(name, health, strength){
+    super(health, strength);
+    this.name = name;
+  }
+  receiveDamage(damage) {
+    this.health = this.health - damage;
+    if (this.health>0) {
+      return `${this.name} has received ${damage} points of damage`;
+    }
+    else if (this.health <=0) {
```
TA-Remote, line 28: here you can write else{} instead of else if ;)

src/viking.js
```diff
@@ -1,14 +1,91 @@
...
+  };
+  addSaxon(saxon) {
+    this.saxonArmy.push(saxon);
+  };
+  vikingAttack() {
+    let randomSaxon = this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
+    let randomViking = this.vikingArmy[Math.floor(Math.random()*this.vikingArmy.length)];
+    let randomSaxonHealth = randomSaxon.health;
+    
+    randomSaxon.receiveDamage(randomViking.strength);
```
TA-Remote, line 74: you need to **return** damage , so you need to return this line .

PR #3090, opened 2022-06-17
Stephanie Tovar - Vikings Solved lab

Stephanie-Tovar:


TA-Remote (commented):
- Code well written: clean, well named and good identation.

- Try to advance in this bonus iteration as it's a good practice also :)

PR #3093, opened 2022-06-17
WD RT JUN 22 | Ruben Andino

rubenanlo:
Still working on the bonuses. Fingers crossed I can find a way to finish this one strong before Monday :)

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,14 +1,86 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor (health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack(){
+    return this.strength;
+  };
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,14 +1,86 @@
...
+    this.vikingArmy.push(viking);
+  }
+
+  addSaxon (saxon) {
+    this.saxonArmy.push(saxon);
+  }
+
+  vikingAttack () {
+    let saxon = saxon.this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)];
+    saxon = new Saxon();
```
TA-Remote, line 64: You don't need to add the Saxons as the tests already does that. And the same goes for vikings :)

PR #3095, opened 2022-06-19
Solved lab

darrenmd:


TA-Remote:
instead of using console.log() , you should use return 

console. log is used for showing data in console. return is used for returning values from a function.

PR #3096, opened 2022-06-19
Solved lab Jabascript vikings Lucas Borchardt 

LucasBorchardt:


src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+  constructor(health, strength){
+    this.health = health;
+    this.strength = strength;
+  }
+  attack(){
+    return this.strength;
+  }
+  receiveDamage(damage){
+    return this.health -= damage
+  }
```
TA-Remote, line 12: According to instructions, receiveDamage() method inside Soldier class should not return anything - only calculate the result

```
  receiveDamage(damage){
    this.health -= damage
  }
```

src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+    return this.strength
+  }
+  receiveDamage(damage){
+    this.health -= damage;
+    if(this.health > 0){
+      return `A Saxon has recived ${damage} points of damage`
+    } else {
+      return "A Saxon has died in combat"
+    }
+}
```
TA-Remote, line 50: Your Saxon class is missing closing curly bracket, that's why tests are crashing ;)

src/viking.js
```diff
@@ -1,11 +1,53 @@
...
-class Saxon {}
+class Saxon extends Soldier {
+  attack(){
+    super.attack();
+    return this.strength
+  }
+  receiveDamage(damage){
+    this.health -= damage;
+    if(this.health > 0){
+      return `A Saxon has recived ${damage} points of damage`
```
TA-Remote, line 46: spelling mistake in word `received`

src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+  attack(){
+    return this.strength;
+  }
+  receiveDamage(damage){
+    this.health -= damage;
+    if(this.health > 0){
+      return `${this.name} has recived ${damage} points of damage` 
+    } else {
+      `${this.name} has died in act of combat"`
+    }
```
TA-Remote, line 30: You must have keyword `return`  in each if else statement. 

If the first condition is falsy, we go into else -> and it is missing `return` now

PR #3097, opened 2022-06-19
Tobias Conrad

maktooob:
good morning :)
had some issues with the return of the attacks


TA-Remote (commented):
- Great Work
- Try to reuse as much code as possible :) 
-  well organized
-  Good names

PR #3098, opened 2022-06-20
Khalid9180

Khalid9180:
I got stuck in class war, I really had no idea what to do next after getting a random variable.
also was not able to do anything in the showStatus method.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,14 +1,83 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength
+  }
+  receiveDamage(damage) {
+this.health = this.health - damage
+  }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor (name, health, strength) {
+    super(health, strength);
+    this.name = name
+  } 
+  receiveDamage(damage) {
+this.health= this.health - damage
+  if (this.health > 0 ) {
+    return `${this.name} has received ${damage} points of damage`
+  } else if (this.health === 0){ 
+return `${this.name} has died in act of combat`
+  }
+  }
+  battleCry() {
+    return "Odin Owns You All!";
+  }
+
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier { 
+  receiveDamage(damage) { 
+    this.health= this.health - damage
+    if (this.health > 0 ) {
+      return `A Saxon has received ${damage} points of damage`
+    } else if (this.health === 0) {
+      return `A Saxon has died in combat`
+    }
+  }
+}
 
 // War
-class War {}
+class War {
+  constructor(){
+    this.vikingArmy = []
+    this.saxonArmy = []
+  }
+  addViking(Viking){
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,14 +1,83 @@
...
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength
+  }
+  receiveDamage(damage) {
+this.health = this.health - damage
```
TA-Remote, line 11: for shorter syntax you could write it like this: 

```
this.health -= damage 
```

src/viking.js
```diff
@@ -1,14 +1,83 @@
...
+  constructor(){
+    this.vikingArmy = []
+    this.saxonArmy = []
+  }
+  addViking(Viking){
+    this.vikingArmy.push(Viking)
+  }
+  addSaxon(Saxon){
+    this.saxonArmy.push(Saxon)
+  }
```
TA-Remote, line 58: i recommend sticking to convention and starting each variable name with lower case ,
`viking` , `saxon` etc

src/viking.js
```diff
@@ -1,14 +1,83 @@
...
+  vikingAttack(){
+    let randomIndex1 = Math.random * this.saxonArmy.length;
+    let index1 = Math.floor(randomIndex1);
+    let randomSaxon = this.saxonArmy[index1];
+    let randomIndex2 = Math.random * this.vikingArmy.length;
+    let index2 = Math.floor(randomIndex2);
+    let randomViking = this.vikingArmy[index2];
+  }
```
TA-Remote, line 66: For the vikingAttack() method you are missing couple steps:

- should make a Saxon receiveDamage() equal to the strength of a Viking

1. Instruction suggests, that the saxon army, that you randomly generated is receiving a damage -> you are calling receiveDamage() method on it

```
randomSaxon.receiveDamage()
```

2. receiveDamage() method is expecting to receive an argument -> you described it in Saxon class.
Instruction suggests the the value of damage is equal to the strength of the viking. If you check attack() method that you described in Soldier class, you can see that this method returns strength of the viking

```
randomSaxon.receiveDamage(randomViking.attack())
```

- should remove dead saxons from the army

For this task you need to find out the index of `randomSaxon` inside `this.saxonArmy` and then remove it from the array

- you would do same tasks for saxonAttack() method

src/viking.js
```diff
@@ -1,14 +1,83 @@
...
+    let randomIndex1 = Math.random * this.saxonArmy.length;
+    let index1 = Math.floor(randomIndex1);
+    let randomSaxon = this.saxonArmy[index1];
+    let randomIndex2 = Math.random * this.vikingArmy.length;
+    let index2 = Math.floor(randomIndex2);
+    let randomViking = this.vikingArmy[index2]
+  }
+  showStatus(){
+
+  }
```
TA-Remote, line 77: this method should only return strings. For example if `this.vikingArmy` is empty array, has no more armies left, you return a strign that saxons won, etc.

TA-Remote (commented):
Good work Khalid! I added suggestions how to continue with the War class. Let me know if still not clear :) 

I will also be uploading solution on student portal

Karina

PR #3113, opened 2022-07-08
[FTLIS072022] Raquel

raqpoletto:


webLisbonTA:
Good job on completing all the mandatory iterations including some bonus, on the last iteration you did you were only missing removing the item from the array

PR #3114, opened 2022-07-08
[FTLISBON072022] Andrea Arestigue

aarestigue:


src/viking.js
```diff
@@ -1,14 +1,115 @@
...
+  
+  vikingAttack() {
+    
+   
+    let resultAttack = (this.saxonArmy[Math.floor(Math.random()*this.saxonArmy.length)]).receiveDamage((this.vikingArmy[Math.floor(Math.random()*this.vikingArmy.length)]).attack());
+    /* for (i=0; i<this.saxonArmy.length; i++){
+      if (this.saxonArmy[i].health <= 0) {
+        this.saxonArmy.splice(this.saxonArmy[i],1);
+        return this.saxonArmy;
+        } */
```
webLisbonTA, line 80: Here you had the perfect logic, the only thing is the "return this.saxonArmy", inside of the loop you shouldn't return anything meaning that your logic was working but you were return the army itself

PR #3145, opened 2022-07-28
Vikings Solved + bonus

avarap:


src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+        // console.log(`saxonAttack 2`);
+        if (viking.health<=0){
+            // console.log(`vikingAttack IN`);
+            this.vikingArmy.pop(vRandom);
+        }
+        return result;
+    }
+    showStatus(){
+        if (this.saxonArmy.length===0) {return "Vikings have won the war of the century!";}
+        else if(this.vikingArmy.length===0) {return "Saxons have fought for their lives and survived another day..."}
```
trebeljahr, line 87: you can drop the else. because if you enter the if block, you'll always return from it, so there is no way how you can get past the if block. functionally they are the same, with that return there and without the else it's called an "early" return pattern. you already do it below by the way, see how there is no else?

src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+    vikingAttack(){
+        // console.log(`vikingAttack 1`);
+        let vRandom=Math.floor(Math.random() * this.vikingArmy.length);
+        let sRandom=Math.floor(Math.random() * this.saxonArmy.length);
+        const viking=this.vikingArmy[vRandom];
+        const saxon=this.saxonArmy[sRandom];
+        let result=saxon.receiveDamage(viking.strength);
+        if (saxon.health<=0){
+            // console.log(`vikingAttack IN`);
+            this.saxonArmy.pop(sRandom);
```
trebeljahr, line 65: this doesn't work - like we talked about in class. pop doesn't accept an argument afaik, so it will just drop the last element and not do what you think it would - namely pop out the element at position sRandom. Use splice, slice, filter or a for loop instead :+1:

src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+        let sRandom=Math.floor(Math.random() * this.saxonArmy.length);
+        const viking=this.vikingArmy[vRandom];
+        const saxon=this.saxonArmy[sRandom];
+        // console.log(`saxonAttack 1`);
+
+        let result=viking.receiveDamage(saxon.strength);
+        // console.log(`saxonAttack 2`);
+        if (viking.health<=0){
+            // console.log(`vikingAttack IN`);
+            this.vikingArmy.pop(vRandom);
```
trebeljahr, line 81: same as above.

PR #3148, opened 2022-07-29
w3 d2 lab

jdlms:


src/viking.js
```diff
@@ -1,11 +1,120 @@
...
+    super(health, strength);
+    this.name = name;
+  }
+
+  //Viking has access to all of "Soldier"'s methods, including "attack()" via "extends"
+
+  //Damage
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (this.health >= 1) {
```
trebeljahr, line 29: could be a ternary as well :)

src/viking.js
```diff
@@ -1,11 +1,120 @@
...
 
 // War
-class War {}
+class War {
+  constructor() {
+    this.vikingArmy = [];
+    this.saxonArmy = [];
+  }
+
+  addViking(Viking) {
```
trebeljahr, line 64: while this works, be careful with the capital letters, since this now "shadows" the name of the Viking class from the same file! also remember, only classes should be named using PascalCase where here in line 64 Viking is an instance of a class not a class itself! so another reason to name it viking instead :)

src/viking.js
```diff
@@ -1,11 +1,120 @@
...
+    this.vikingArmy.push(Viking);
+  }
+
+  addSaxon(Saxon) {
```
trebeljahr, line 68: same here.

src/viking.js
```diff
@@ -1,11 +1,120 @@
...
+    this.saxonArmy.push(Saxon);
+  }
+
+  vikingAttack() {
+    let rViking = Math.floor(Math.random() * this.vikingArmy.length);
+    let rSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+    let attack = this.vikingArmy[rViking].strength;
+    this.saxonArmy[rSaxon].receiveDamage(attack);
```
trebeljahr, line 76: save the return of the call to the receiveDamage function in a variable and then return the value of that variable after the splice.
this way you can also get rid of line 79! :)

src/viking.js
```diff
@@ -1,11 +1,120 @@
...
+    // let damage = rViking.strength;
+    // let rSaxon = this.random(saxonArmy);
+    // rSaxon.receiveDamage(damage);
+
+    // let oldHealth = rSaxon.health;
+    // rSaxon.health = oldHealth -= rViking.strength;
+  }
+
+  saxonAttack() {
+    let rViking = Math.floor(Math.random() * this.vikingArmy.length);
```
trebeljahr, line 94: the Math.random stuff you have used 4 times now, some refactoring might be in order :)

src/viking.js
```diff
@@ -1,11 +1,120 @@
...
+    let rSaxon = Math.floor(Math.random() * this.saxonArmy.length);
+    let attack = this.saxonArmy[rSaxon].strength;
+    this.vikingArmy[rViking].receiveDamage(attack);
```
trebeljahr, line 97: same as in the comment above.

trebeljahr (commented):
good work :) a few little things with the bonus iterations but still really good job :)

PR #3149, opened 2022-07-29
[PTBER072022] berndHavekost

hvkst:


src/viking.js
```diff
@@ -1,11 +1,94 @@
...
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (this.health > 0) {
```
trebeljahr, line 24: ternary here could be cool :)

src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+    return "Odin Owns You All!";
+  }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  receiveDamage(damage) {
+    this.health -= damage;
+    if (this.health > 0) {
```
trebeljahr, line 38: same here.

src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+
+  vikingAttack() {
+    return this.attack(this.vikingArmy, this.saxonArmy);
+  }
+
+  saxonAttack() {
+    return this.attack(this.saxonArmy, this.vikingArmy);
+  }
+
+  removeDeadSoldier(whatArmy) {
```
trebeljahr, line 67: this is a neat little function. :) 
you will love filter, because it does that but much easier. also this might not behave as you think it would in some edge cases, because you change the length of the array and shift around indeces as you do so, so generally when you mutate an array, while you are still iterating over it, that can lead to subtle bugs that are hard to catch.

src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+    whatArmy.forEach((soldier, index) => {
+      if (soldier.health < 1) whatArmy.splice(index, 1);
+    });
+  }
+
+  attack(offense, defense) {
```
trebeljahr, line 73: super cool refactor :) pretty much what we had in class :+1:

src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+    const offSoldier = this.random(offense);
+    const defSoldier = this.random(defense);
+    const fight = defSoldier.receiveDamage(offSoldier.strength);
+    this.removeDeadSoldier(defense);
+    return fight;
+  }
+
+  showStatus() {
+    if (this.saxonArmy.length === 0) {
```
trebeljahr, line 82: you can drop the elses down there. 
it's called "early" return pattern and most people think it's easier to read than nested if elses. you might even be able to inline some of the returns and not give them their own {} code blocks :)

trebeljahr (commented):
really good work, I like the cleanliness of the code and that it's refactored already to a really good extent! keep it up :)

PR #3150, opened 2022-07-29
Solved lab

herrFernandez:
Done! ( I got me some help for the attack function though. Such a good friend :-D )

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,102 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+
+    attack() {
+        return this.strength = this.strength;
+      }
+
+    receiveDamage(damage) {
+        this.health = this.health - damage;
+      }
+
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength)  {
+        super(health, strength);
+        this.name = name;
+    }
+    
+    receiveDamage(damage) {
+        this.health = this.health - damage;
+        if (this.health > 0) {
+            return `${this.name} has received ${damage} points of damage`;
+        } else {
+            return `${this.name} has died in act of combat`;
+        }
+      }
+    
+    battleCry () {
+        return "Odin Owns You All!";
+    }  
+}
+
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    constructor(health, strength)  {
+        super(health, strength);
+    }
+
+    receiveDamage(damage) {
+        this.health = this.health - damage;
+        if (this.health > 0) {
+            return `A Saxon has received ${damage} points of damage`;
+        } else {
+            return `A Saxon has died in combat`;
+        }
+      }
+    
+}
 
 // War
-class War {}
+class War {
+    constructor() {
+        this.vikingArmy = [];
+        this.saxonArmy = [];
+    }
+
+    addViking(viking) {
+        this.Viking = viking;
+        this.vikingArmy.push(this.Viking);
+    } 
+    addSaxon(saxon) {
+        this.Saxon = saxon;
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,102 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+
+    attack() {
+        return this.strength = this.strength;
```
trebeljahr, line 9: why the assignment to itself? 
you could just do return this.strength :man_shrugging: and it would do the exact same and be shorter :)

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+      }
+
+    receiveDamage(damage) {
+        this.health = this.health - damage;
```
trebeljahr, line 13: can be shortened to this.health -= damage;

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength)  {
+        super(health, strength);
+        this.name = name;
+    }
+    
+    receiveDamage(damage) {
+        this.health = this.health - damage;
```
trebeljahr, line 26: same.

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+        if (this.health > 0) {
```
trebeljahr, line 27: could use a ternary here instead of the if/else returns. 
return this.health ? `${this.name} has received ${damage} points of damage` : `${this.name} has died in act of combat`

both work, just so you know :)

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+        this.vikingArmy = [];
+        this.saxonArmy = [];
+    }
+
+    addViking(viking) {
+        this.Viking = viking;
+        this.vikingArmy.push(this.Viking);
+    } 
+    addSaxon(saxon) {
+        this.Saxon = saxon;
```
trebeljahr, line 69: this is not necessary. 
you can just add the argument directly like so: 
this.vikingArmy.push(viking) 

No need for another property on the War class object to be added.

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+        this.saxonArmy.push(this.Saxon);
+    } 
+
+    vikingAttack() {
+        const damageResult = this.Saxon.receiveDamage(this.Viking.strength);
+        
+        this.saxonArmy = this.saxonArmy.filter(saxon => saxon.health > 0);
```
trebeljahr, line 76: haha nice filter function. Do you understand what this does, because we haven't done this in class yet?

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+      
+      return damageResult;
+    }
+
+    saxonAttack() {
+        const damageResult = this.Viking.receiveDamage(this.Saxon.strength);
+        
+        this.vikingArmy = this.vikingArmy.filter(viking => viking.health > 0);
```
trebeljahr, line 84: you could split out the filtering functionality. also this doesn't pick a random soldier. 

it passes the test, because there is only one soldier added but the way this is written the last soldier added to the arrays of the army will be the ones fighting. which is not random at all!

trebeljahr (commented):
good work, there is one of the comments I would like you to think about some more, namely that where you assign this.Viking = viking part. It's doing something that's not quite what we want there and I want to make sure that you understand what is going on :)

src/viking.js
herrFernandez: filter saxon and leave only the ones tht are not dead (Yet :-D )

src/viking.js
herrFernandez: true! i missed the ramdom part (I have to read the exercises more carefully) I'll try to do it this week :-)

PR #3369, opened 2022-11-26
[FTRMT112022] Jonnathan

Jogopin:


Jogopin:
the new method for bonus 5 should work( unless I have some typo issues), however I`m curios if there is a  way to create an undefined matrix where i can push new elements into new arrays into the matrix. 
the way i solved  was  creating  a matrix of 3 empty arrays, but that comes with the problem that if i want to add a new army to the game i have to add a new  empty array  in  this.allTheArmies in the constructor of the class War, and my intention was to avoid that issue by adding a new array to the matrix every time i create a new method to add a new faction.

PR #3436, opened 2023-01-20
[LIS-FT-01-23] Duarte França

Duartefranca:


src/viking.js
```diff
@@ -1,11 +1,89 @@
...
+        receiveDamage(damage) {
+         this.health -= damage;
+      }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength) {
+        super(name, health, strength)
```
webLisbonTA, line 18: Viking is extending the class Soldier, this means that Viking inherit health and strength from Soldier.
```
    super(health, strength)
    this.name = name
}
(...)
```

src/viking.js
```diff
@@ -1,11 +1,89 @@
...
+  }
+
+
+  vikingAttack() {
+    let saxonIndex = Math.floor(Math.random() * this.saxonArmy.length)
+    let vikingIndex = Math.floor(Math.random() * this.vikingArmy.length)
+    
+    let saxonAttaked = this.saxonArmy[saxonIndex].receiveDamage(this.vikingArmy[vikingIndex].strength)
+    if (this.saxonArmy[saxonIndex].health <= 0)  this.saxonArmy.pop(saxonIndex); 
+      return saxonAttaked
```
webLisbonTA, line 74: Here, you should've used another array method like `.splice()`, just because the `.pop()` method removes the last element of the array. Even if you pass one index inside the method `.pop()`, it will always remove the last element of the array. It passes the tests because it's jasmine, but in another case it probably wouldn't work the way it is supposed to. 
The overall logic is correct, so, here's a method you could use: 
 ```
if (this.saxonArmy[saxonIndex].health <= 0) {
     this.saxonArmy.splice([saxonIndex], 1);
      }
```

webLisbonTA (commented):
Good job on the lab Duarte. Just check the annotations I left you for a better understanding.

PR #3441, opened 2023-01-20
solved without bonus (yetttt)

miacsilva:


src/viking.js
```diff
@@ -1,11 +1,71 @@
...
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength) {
+        super(health, strength)
+        this.name = name
+    }
+    attack() {
```
joaopdg, line 21: Since `Viking` is extending the class `Soldier` you don't need to declare the `attack()` method here again because it inherits it from the `Soldier:`

```
    super(health, strength);
    this.name = name;
  }
  receiveDamage(theDamage) {
    this.health -= theDamage;
    if (this.health > 0)
      return `${this.name} has received ${theDamage} points of damage`;
  }
  battleCry() {
    return "Odin Owns You All!";
  }
}
```

src/viking.js
```diff
@@ -1,11 +1,71 @@
...
+    }
+}
+
+
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    constructor(health, strength) {
+        super(health, strength)
```
joaopdg, line 40: Since `Saxon` is extending the class `Soldier` and the only thing that is different in the `Saxon` class is the `receiveDamage()` method you don't need the `constructor` neither the `super` here:

```
class Saxon extends Soldier {
  receiveDamage(theDamage) {
    this.health -= theDamage;
    if (this.health > 0)
      return `A Saxon has received ${theDamage} points of damage`;
  }
}
```

src/viking.js
```diff
@@ -43,29 +41,43 @@ class Saxon extends Soldier {
...
-        }
+        if (this.health> 0) return `A Saxon has received ${theDamage} points of damage`
+        
+        else return `A Saxon has died in combat`
+        
 } }
 
 class War {
-    constructor()
+    vikingArmy = []
```
joaopdg, line 51: When declaring classes you always should put `this.` before the keys:

```
class War {
  constructor() {
    this.vikingArmy = [];
    this.saxonArmy = [];
  }

(...)
```

src/viking.js
```diff
@@ -43,29 +41,43 @@ class Saxon extends Soldier {
...
+    saxonArmy = []
+
+    addViking(viking){
+    this.vikingArmy.push(viking)
+    }
+    addSaxon(saxon){
+    this.saxonArmy.push(saxon)
+    }
+    vikingAttack(){
```
joaopdg, line 60: This iteration is not passing this teste `"should return result of calling receiveDamage() of a Saxon with the strength of a Viking"`.
This is happening because you are returning the `this.saxonArmy.splice(saxon, 1);` inside the `if`, and the result of the attack is being returned inside an `else`. 
This means that the code will run and if `(!(this.saxonArmy[saxon].health > 0))` is true you will remove the dead saxon from the array and it will stop there, so the `return valor` will never be read.
To solve this you should do something like this:

```
(...)

  vikingAttack() {
    let viking = Math.floor(Math.random() * this.vikingArmy.length);
    let saxon = Math.floor(Math.random() * this.saxonArmy.length);
    let valor = this.saxonArmy[saxon].receiveDamage(
      this.vikingArmy[viking].strength
    );

    if (!(this.saxonArmy[saxon].health > 0)) {
      this.saxonArmy.splice(saxon, 1);
    }

    return valor;
  }

(...)
```

joaopdg (commented):
Great job on the lab Maria!

PR #3444, opened 2023-01-20
[LISFT0123]João Raposo

JRaposo22:


src/viking.js
```diff
@@ -1,11 +1,116 @@
...
+
+    vikingAttack(){
+
+        let randomViking = Math.floor(Math.random() * this.vikingArmy.length);
+        let randomSaxon =  Math.floor(Math.random() * this.saxonArmy.length);
+
+
+        let attackResult = this.saxonArmy[randomSaxon].receiveDamage(this.vikingArmy[randomViking].strength);
+
+        if(this.saxonArmy[randomSaxon].health <= 0) this.saxonArmy.pop(randomSaxon);
```
joaopdg, line 83: The array method `.pop()` will exclude only the last index of the array. Since the Saxon will always be random, and we don't know if it will be the first or the last index of the array, the correct method to use here is `.splice()` because it let's you exclude an element from any index of the array. The method `.splice()` accepts two arguments: the index from where you want to start to delete elements, and how many elements you want to delete:

```
(...)

    if (this.saxonArmy[randomSaxon].health <= 0)

(...)
```

src/viking.js
```diff
@@ -1,11 +1,116 @@
...
+    
+    }
+
+    saxonAttack(){
+        let randomViking = Math.floor(Math.random() * this.vikingArmy.length);
+        let randomSaxon =  Math.floor(Math.random() * this.saxonArmy.length);
+
+        let attackResult = this.vikingArmy[randomViking].receiveDamage(this.saxonArmy[randomSaxon].strength);
+
+        if(this.vikingArmy[randomViking].health <= 0) this.vikingArmy.pop(randomViking);
```
joaopdg, line 97: Same thing here:

```
(...)

    if (this.vikingArmy[randomViking].health <= 0)
      this.vikingArmy.splice(randomViking, 1);

(...)
```

joaopdg (commented):
Great job on the lab João!

PR #3445, opened 2023-01-20
[LISFT0123] Alexandre Álvaro

AleAlvo:
- lab done with bonus

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,98 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+
+    attack() {
+        return this.strength;
+    }
+
+    receiveDamage(damage) {
+        this.health -= damage;
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor (name, health, strength){
+        super(health, strength);
+        this.name = name;
+    }
+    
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+    addSaxon(Saxon){
+        this.saxonArmy.push(Saxon);
+    }
+    
+    vikingAttack(){
+        let saxonIndex = Math.floor(Math.random() * this.saxonArmy.length);
+        let vikingIndex = Math.floor(Math.random() * this.vikingArmy.length);
+        let attackResult = this.saxonArmy[saxonIndex].receiveDamage(this.vikingArmy[vikingIndex].strength);
+        if (this.saxonArmy[saxonIndex].health <= 0) {
+            this.saxonArmy.pop(saxonIndex)
```
webLisbonTA, line 67: Here, you should've used another array method like `.splice()`, just because the `.pop()` method removes the last element of the array. Even if you pass one index inside the method `.pop()`, it will always remove the last element of the array. It passes the tests because it's jasmine, but in another case it probably wouldn't work the way it is supposed to. 
The overall logic is correct, so, here's a method you could use: 
 ```
if (this.saxonArmy[saxonIndex].health <= 0) {
     this.saxonArmy.splice([saxonIndex], 1);
      }
```

webLisbonTA (commented):
Amazing job on the lab Alexandre. Please check out the notes I left you.

PR #3449, opened 2023-01-20
[LISFT0123] Gabriel

gbrlcsgn:


src/viking.js
```diff
@@ -1,11 +1,53 @@
...
+    }
+    battleCry(){
+        return `Odin Owns You All!`
+    }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    attack() {
```
joaopdg, line 39: Since `Saxon` is extending the class `Soldier` and the only thing that is different in the `Saxon` class is the `receiveDamage()` method you don't need the `constructor` neither the `super` here:

```
class Saxon extends Soldier {
  receiveDamage(damage) {
    this.health -= damage;
    if (this.health > 0)
  }
}
```

joaopdg (commented):
Great job on the lab Gabriel, next time try to do the bonus!

PR #3455, opened 2023-01-20
[FT0123] Déborah 

deborahdelimaa:
couldn't solve the bonus one.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,63 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength){
+        this.health= health;
+        this.strength = strength;
+    }
+    attack() {
+    return this.strength;
+    }
+    receiveDamage(damage){
+        this.health = this.health - damage;
+    }
+}
 
-// Viking
-class Viking {}
+// Viking 
+class Viking extends Soldier{
+    constructor(name, health, strength){
+        super(health,strength) 
+            this.name = name;
+    }
+attack(){
+    return this.strength;
+}
+receiveDamage(damage){
+    this.health -= damage;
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+    this.saxonArmy = [];
+   }
+   addViking(axe) {
+    this.vikingArmy.push(axe);
+   }
+   addSaxon(key){
+    this.saxonArmy.push(key);
+   }
+   vikingAttack() {
+    let random = Math.random().Viking;
```
webLisbonTA, line 61: A solution you could've done is:
```
vikingAttack() {
        let randomVikingIndex = Math.floor(Math.random() * this.vikingArmy.length);
        let result = this.saxonArmy[randomSaxonIndex].receiveDamage(this.vikingArmy[randomVikingIndex].strength);
        if (this.saxonArmy[randomSaxonIndex].health <= 0) {
        this.saxonArmy.splice([randomSaxonIndex], 1)}
        return result;
    }
```
First, you store the value of and random index for the viking and the saxon, by using the `Math.floor(Math.random() * this.saxonArmy.length);` method, and multiplying it for the length of the array, so the random value always stays inside the length of the array we want to randomize.
After that, we initialize one variable that will store the result of the fight we are creating. In this 'fight', we grab the random saxon index that we created, and use the `receiveDamage()` function that we had already created earlier. We pass the strength of a random viking and then check the results of that by using a conditional that checks if the saxon stayed alive or died. In the case of him dying, we removed him from the original array.

webLisbonTA (commented):
Good job on the lab. Please check the comment reviews for a better understanding.

PR #3458, opened 2023-01-20
[LISFT0123]Farid

Farid-97:


src/viking.js
```diff
@@ -1,11 +1,89 @@
...
+    }
+    receiveDamage(damage){
+        this.health-=damage
+    }  
+}
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name,health,strength){
+    super(name,health,strength)
```
joaopdg, line 17: Viking is extending the class Soldier, this means that Viking inherit health and strength from Soldier.

```
    super(health, strength)
    this.name = name
}

(...)
```

In the constructor you need to pass the 3 arguments, in the super you only need to pass health and strength, and then you just need to declare the variable name, since health and strength is already declared at the class Soldier.

src/viking.js
```diff
@@ -1,11 +1,89 @@
...
+        else {
+    return(`A Saxon has died in combat`)
+        }
+    }
+}
 
 // War
-class War {}
+class War {
+    vikingArmy=[]
```
joaopdg, line 52: When declaring classes you always should put `this.` before the keys:

```
class War {
  constructor() {
    this.vikingArmy = [];
    this.saxonArmy = [];
  }

(...)
```

src/viking.js
```diff
@@ -1,11 +1,89 @@
...
+return fight2
+}
+showStatus(){
+    if (this.saxonArmy<=0){
+        return "Vikings have won the war of the century!"
+    }
+    else if (this.vikingArmy<=0) {
+return "Saxons have fought for their lives and survived another day..."
+    }
+    else if(this.saxonArmy=this.vikingArmy){
```
joaopdg, line 85: Here you want to compare the vikingArray with the saxonArray but you're doing it with just one `=` this is used to reassign a new value to a variable, if you want to compare something you need to use `==` or `===`
```
(...)

    else if(this.saxonArmy === this.vikingArmy){
        return "Vikings and Saxons are still in the thick of battle."
    }

(...)
```
But this is not what you are asked to do in the lab instructions. They want you to print this message if there are still vikings and saxons alive, so the correct condition is:
```
(...)

else if (this.saxonArmy.length >= 1 && this.vikingArmy.length >= 1) {
    }

(...)
```

joaopdg (commented):
Great job on the lab Farid!

PR #3460, opened 2023-01-20
[LISFT0123] André Lisboa

andrelisboapt:
lab done with bonus

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,97 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength){
+        this.health = health;
+        this.strength = strength;
+        
+    }
+    attack(){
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+        
+    }
+}
+
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength){
+        super(name, strength)
```
webLisbonTA, line 21: Since you are extending the Soldier class, you should pass "health" and "strength" in the super, instead of the name.
When referencing you only needed to do `this.name = name;` because the other values you are extending from the parent class Soldier.

webLisbonTA (commented):
Nice work on the lab André. Please check out the notes.

PR #3477, opened 2023-01-22
done

evelinecoen:


src/viking.js
```diff
@@ -1,11 +1,64 @@
...
-class Viking {}
+class Viking extends Soldier {
+    constructor (name, health, strength) {
+    super (health, strength) // the ones that are in the "mother" class 
+    this.name = name 
+    }
+    receiveDamage(damage){
+       this.health = (this.health - damage)    // we need to repeat this one (and not the attack()etc because the conditions are part of this step
+       if (this.health > 0) return `${this.name} has received ${damage} points of damage`
+       if (this.health === 0) return `${this.name} has died in act of combat`
```
webLisbonTA, line 24: Instead of another `if` statement, you should use the `else`, because the health can be less than 0, if the attack is much bigger than the health.
```
if ( this.health > 0 ) return `${this.name} has received ${damage} points of damage`
```
It's better to use the `else`, because it will take care of every event that doesn't match that `if` condition above.

src/viking.js
```diff
@@ -1,11 +1,64 @@
...
+    this.saxonArmy = [];
+}
+ addViking (vikingObject) {
+    this.vikingArmy.push (vikingObject);
+ }
+ addSaxon(saxonObject) {
+    this.saxonArmy.push (saxonObject);
+ }
+vikingAttack(){ 
+}
```
webLisbonTA, line 54: A solution you could've done is:
```
vikingAttack() {
        let randomVikingIndex = Math.floor(Math.random() * this.vikingArmy.length);
        let result = this.saxonArmy[randomSaxonIndex].receiveDamage(this.vikingArmy[randomVikingIndex].strength);
        if (this.saxonArmy[randomSaxonIndex].health <= 0) {
        this.saxonArmy.splice([randomSaxonIndex], 1)}
        return result;
    }
```
First, you store the value of and random index for the viking and the saxon, by using the `Math.floor(Math.random() * this.saxonArmy.length);` method, and multiplying it for the length of the array, so the random value always stays inside the length of the array we want to randomize.
After that, we initialize one variable that will store the result of the fight we are creating. In this 'fight', we grab the random saxon index that we created, and use the `receiveDamage()` function that we had already created earlier. We pass the strength of a random viking and then check the results of that by using a conditional that checks if the saxon stayed alive or died. In the case of him dying, we removed him from the original array.

webLisbonTA (commented):
Good job on the lab. Check the annotations I left you for a better understanding. 
Next time, try to do the bonuses.

PR #3479, opened 2023-01-22
[LISFT0123] Rafaela Umbelino

RafaelaUmbelino:


src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+    }
+    addViking(Viking){
+        this.vikingArmy.push(Viking)
+
+    }
+
+    addSaxon(Saxon){       
+        this.saxonArmy.push(Saxon)
+    }
+    vikingAttack(){
```
lzaquine, line 59: Here, the vikingAttack() function isn't working because you're not properly referencing the Viking army. Also, it uses Viking.strength to attack the Saxon, instead of getting a random Viking from the Viking army and using its attack method. And you're not calling splice correctly. It should be called on the saxonArmy array and passing in the index of the Saxon to remove and the number of items to remove (1) as arguments, not indexSaxon and indexSaxon.

This should fix it:

```
vikingAttack() {
        let indexViking = Math.floor(Math.random() * this.vikingArmy.length);
        let randomViking = this.vikingArmy[indexViking];
        let indexSaxon = Math.floor(Math.random() * this.saxonArmy.length);
        let randomSaxon = this.saxonArmy[indexSaxon];
        let saxonDamage = randomSaxon.receiveDamage(
          randomViking.attack()
        );
          this.saxonArmy.splice(indexSaxon, 1);
        }
        return saxonDamage;
      }
```

src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+        if(randomSaxon.health <= 0){
+            this.saxonArmy.length.splice(indexSaxon, indexSaxon)
+        }
+
+        return saxonDamage;
+
+
+        
+    }
+    saxonAttack(){
```
lzaquine, line 74: Same as explained before!
This should fix it:

```
saxonAttack() {
        let vikingIndex = Math.floor(Math.random() * this.vikingArmy.length);
        let randomViking = this.vikingArmy[vikingIndex];
        let saxonIndex = Math.floor(Math.random() * this.saxonArmy.length);
        let randomSaxon = this.saxonArmy[saxonIndex];
        let vikingDamage = randomViking.receiveDamage(
          randomSaxon.attack()
        );
        if (randomViking.health <= 0) {
          this.vikingArmy.splice(vikingIndex, 1);
        }
        return vikingDamage;
      }
```

lzaquine (commented):
Great job on the lab, Rafaela!

PR #3481, opened 2023-01-23
[LISFT0123] Assebe Kleinert

Assebe:


src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+    soldier1.receiveDamage(50)
+    console.log(soldier1)
+    
+    
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength){
+    super(health, strength)
+    this.name = name
```
webLisbonTA, line 26: Viking is extending the class Soldier, this means that Viking inherit health and strength from Soldier.
```
    super(health, strength)
    this.name = name
}
(...)
```

src/viking.js
```diff
@@ -1,11 +1,63 @@
...
+    this.health = health
+    this.strength = strength
+
+    }
+    receiveDamage(damage){
+        this.health - damage
```
webLisbonTA, line 32: Here, you need to reference the `this.health`, being it's value minus the damage, so, you need to store that value:
```
receiveDamage(damage) {
     this.health -= damage
     if (this.health > 0) {
     (...) }
```

webLisbonTA (commented):
Good job on the lab. Check the notes for a better understanding, and next time, try to do the bonuses.

PR #3518, opened 2023-02-25
[RMT-FT-FEB-2023] Paulius Mockevicius

pmockevicius:
Hi there....

I found this lab not so much educational but more like a test of how well you follow directions. Every step was given in the description and just by reading carefully them i was able to finish all iterations but still i don't feel like i fully understand classes and will be doing more research during the weekend....

Looking forward to learn more about classes. I find them really interesting

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -61,14 +61,43 @@ class War {constructor(){
         
     }
     vikingAttack(){
-        console.log(this.vikingArmy[0].strength)
-        // Saxon.receiveDamage(this.vikingArmy[0].strength)
+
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Paulius! 
Great job on this lab, the overall structure is real nice and clean!
One small improvement could be on the **_randomViking_** or **_randomSaxon_**  index. You could use the _**vikings/saxons**_ array.length to generate a number that will fit the array to avoid nonexistent index. 
Other than that, amazing work, keep it up!

PR #3526, opened 2023-02-25
[RMT.FT.WD.FEB-23] Sotiris Ntokas

SotDok:
Couldn t solve the bonus.Having trouble understunding how I need to use what I already have over there.

maxresdefault

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,94 @@
 // Soldier
-class Soldier {}
+class Soldier{
+  constructor(health, strength) {
+  this.health = health;
+  this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+
+  receiveDamage(damage){
+    this.damage = damage;
+    this.health = this.health - damage;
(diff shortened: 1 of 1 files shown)


TA-Remote:
HIIIII!
Often you'll be asked to pick a random element from an array. And as you mentioned, we'll use the Math.random()... -> this method return a number between 0 and almost 1
so, we want a random viking from the viking array and  to get that element we need the index
1. index needs to be random.
2. index needs to be an integer --> we can use [Math.floor()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/floor) 
3. index between 0 and the leght of our array --> if for example, our array has 3 element, should be index 0, 1, or 2.
4. const randomVikingIndex = Math.floor(Math.random() * this.vikingArmy.length))
5. then we can have our single player from the array--> const randomViking = this.vikingArmy[randomVikingIndex]
Hope it can help😉

PR #3546, opened 2023-02-26
[rmtft20.feb.2023]Peng Chen

Mr-Chennn:
done


ykztenoks:
Hey Peng! 
Great work on this lab, nice structure and clean code!
One thing I could say is that you could create a new instance of viking/saxon class outside of the attack method/function, just to make it cleaner, but for testing purposes is no problem at all! great job!

PR #3547, opened 2023-02-26
[rmt wd ft 0223]Pim

8lom:
stranded at war attack


TA-Remote:
hey Pim!
whenever you wanna add an element to an array we use the .push() method -->  array.push(element)
Here just a cute way to remember some methods😉
pop-shift-unshift-push
🙌

PR #3550, opened 2023-02-27
[FTRMT022023] Hugo

hugowannabedev:
I couldn't understand how to "add the received Viking to the army" for the bonus, iteration 4: War.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -28,7 +28,7 @@ class Viking extends Soldier {
     }
     receiveDamage(damage) {
         this.health = this.health - damage;
-        if (this.health > damage) {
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Hugo!
when working on array (and you'll do it a looooot) there are some methods that you are gonna use a looooot, are just the more common! (here just a cute rapresentation😉)
pop-shift-unshift-push 
Another method to remove a specific element from an array is  [.splice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) take a look at the documentation when you have time, and if yuo have any other questions let us know.🙌

PR #3552, opened 2023-02-27
[RMT FT WD 022023] Mattie Barton

maitlandbarton:
completed lab + attempted bonus..could not get past armies attack!

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,77 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+    attack(){
+        return this.strength;
+    }
+    receiveDamage(damage){
+        this.health -= damage;
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+      super(health, strength);
+      this.name = name;
+    }
+    receiveDamage(damage){
+      this.health -= damage;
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Mattie! 
you were on the right path using th .slice() methods...
Take also a look at the documentation here: [.spice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)
As you'll see you can pass more parameters to this methods, but let focus on what we actually need
`
// example: remove 1 element at index 3
const myFish = ["angel", "clown", "drum", "mandarin", "sturgeon"];
const removed = myFish.splice(3, 1);

// myFish is ["angel", "clown", "drum", "sturgeon"]
// removed is ["mandarin"]`
The first parameter is the index --> the position in the array of the element you want to remove (from where to start)
The second parameter is the numbers of element you want to remove.
`
vikingAttack() {
    
      // store randomVikingIndex --> generate a random index from VikingArmy
      // store the rondomViking --> this.vikingArmy[randomVikingIndex];
      // do the same for the Saxon
     
       const result = randomSaxon.receiveDamage(randomViking.attack()); --> the attack method return the strenght 
  
        this.saxonArmy.splice( randomSaxonIndex   , 1); --> use the store random index to remove
        return ("A Saxon has died in combat");
      } else if 
      (randomSaxon.health > 0) {
        return (`A Saxon has received ${damage} points of damange`);
      }
    }
`
Great job🙌

PR #3608, opened 2023-04-14
LIS-WD-FT-APR-23-TomásMaria

SinTomas:


src/viking.js
```diff
@@ -1,11 +1,97 @@
...
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health,strength);
+        this.name=name;
+    };
+    attack() {
+        return this.strength;
```

src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    constructor(health, strength) {
+        super(health,strength);
+    };
+    attack() {
+        return this.strength;
```
joaopdg, line 43: Same thing as before, if the method is the same, there's no need to repeat it

src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+    };
+    addSaxon(saxon){
+        this.saxonArmy.push(saxon);
+    };
+    vikingAttack(){
+        let randomSaxon = Math.floor(Math.random()*this.saxonArmy.length);
+        let randomViking = Math.floor(Math.random()*this.vikingArmy.length);
+        let result = this.saxonArmy[randomSaxon].receiveDamage(this.vikingArmy[randomViking].strength);
+        if (this.saxonArmy[randomSaxon].health<=0) {
+            this.saxonArmy.pop(randomSaxon);
```
joaopdg, line 73: .pop() always removes the last element of the array, this is passing the test but you are not removing the Saxon that received the damage, for that you should have used the .splice()

src/viking.js
```diff
@@ -1,11 +1,97 @@
...
+        };
+        return result;
+    };
+    saxonAttack(){
+        let randomViking = Math.floor(Math.random()*this.vikingArmy.length);
+        let randomSaxon = Math.floor(Math.random()*this.saxonArmy.length);
+        let result = this.vikingArmy[randomViking].receiveDamage(this.saxonArmy[randomSaxon].strength);
+        if (this.vikingArmy[randomViking].health<=0) {
+            this.vikingArmy.pop(randomViking);
```
joaopdg, line 82: Same thing as before, use .splice() instead of .pop()

joaopdg (commented):
Great job on the lab and the bonus Tomás!

PR #3609, opened 2023-04-14
Solved lab

k-129:


src/viking.js
```diff
@@ -1,11 +1,101 @@
...
+        this.saxonArmy.push(saxonObject);
+
+    };
+    vikingAttack() {
+        let randomSaxon = Math.floor(Math.random()*this.saxonArmy.length);
+        let randomViking = Math.floor(Math.random()*this.vikingArmy.length);
+        let result = this.saxonArmy[randomSaxon].receiveDamage(this.vikingArmy[randomViking].strength)
+
+        if (this.saxonArmy[randomSaxon].health <= 0) {
+            this.saxonArmy.pop(randomSaxon);
```
brunorocha20, line 77: This `.pop()` method, always removes the last index from the array. Even you passing the random saxon index inside, it will always remove the last index. You are passing the jasmine tests, but it is not correct, as you are always removing the last saxon from the `saxonArmy` array.

src/viking.js
```diff
@@ -1,11 +1,101 @@
...
+        let result = this.vikingArmy[randomSaxon].receiveDamage(this.saxonArmy[randomSaxon].strength)
+
+        if (this.vikingArmy[randomViking].health <= 0) {
+            this.vikingArmy.splice(randomViking,1);
+        }
+        return result;    
+    };
+
+    showStatus() {
+        if (this.saxonArmy == 0) {
```
brunorocha20, line 93: Don't forget to always use the `===` instead of just `==`, this time it's ok, but later in the code, it can lead to bugs and/or errors.

brunorocha20 (commented):
Great job on the lab Erik.

PR #3610, opened 2023-04-14
[LIS-WDFT-APR23] Joao Madeira

jodmadeira:


src/viking.js
```diff
@@ -1,11 +1,104 @@
...
+        super(health, strength)
+        this.name = name
+    };
+    attack(){
+        return this.strength
+    }
+    receiveDamage(damage) {
+        this.health=this.health-damage
+        if(this.health>0){
+            return this.name + ` has received ` + damage + ` points of damage`
```
brunorocha20, line 27: Here, you can use `string interpolation` to return the message.

brunorocha20 (commented):
Great job on the lab, João.

PR #3612, opened 2023-04-14
[WD FT APR2023] Ana Rak

ana-rak:


src/viking.js
```diff
@@ -1,11 +1,96 @@
...
+      this.saxonArmy[randomSaxon].strength
+    );
+
+    if (this.vikingArmy[randomViking].health <= 0) {
+      this.vikingArmy.pop(randomViking);
+    }
+    return result;
+  }
+  showStatus() {
+    if (this.saxonArmy == 0) {
```
brunorocha20, line 88: Always remember to use the three equal signs `===` instead of just `==`. In this case it's ok, but can lead to mistakes and/or bugs later in the code.

src/viking.js
```diff
@@ -1,11 +1,96 @@
...
+  battleCry() {
+    return "Odin Owns You All!";
+  }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+  constructor(health, strength) {
+    super(health, strength);
```
brunorocha20, line 37: Since the `Saxon` class extends `Soldier`, and you don't need to add anything to the `constructor`, you can "skip" the `constructor` and `super` and move on the the methods you want to add.

brunorocha20 (commented):
Great job on the lab Ana.

PR #3621, opened 2023-04-14
[LIS-WD-FT-APR-2023] Margarida Pereira

mapmargaridapereira:


src/viking.js
```diff
@@ -1,11 +1,77 @@
...
+  }
+  battleCry() {
+    return "Odin Owns You All!";
+  }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    constructor(health, strength, name) {
```
joaopdg, line 36: Saxon will not receive a name so there's no need for the constructor neither the super, since it's inheriting everything from the Soldier class:

```
class Saxon extends Soldier {
  receiveDamage(damage) {
    this.health -= damage;

    if (this.health <= 0) {
    } else {
    }
  }
}
```

joaopdg (commented):
Great job on the lab and the bonus Margarida!

PR #3628, opened 2023-04-14
[lis-wd-ft-apr]Niroj

Niroj167:


src/viking.js
```diff
@@ -28,11 +28,11 @@ class Soldier {
       if(this.health <= 0 ){ 
           return `${this.name} has died in act of combat`;
      } else {
-         return `${this.name} has received ${Damage} points of damage`;
+         return `${this.name}has received ${Damage} points of damage`;
```
joaopdg, line 31: failing this test "should return "NAME has received DAMAGE points of damage", if the Viking is still alive" because you forgot to put a space between "this.name" and "has":
```
return `${this.name} has received ${Damage} points of damage`;
```

src/viking.js
```diff
@@ -44,7 +44,8 @@ class Soldier {
     super.receiveDamage(Damage);
     if(this.health > 0 ){ 
         return `A Saxon has received  ${Damage} points of damage`;
```
joaopdg, line 46: failing this test "should return "A Saxon has received DAMAGE points of damage", if the Saxon is still alive" because you have 2 spaces between "received" and "Damage" and you should have one:
```
return `A Saxon has received ${Damage} points of damage`;
```

joaopdg (commented):
Great job on the lab Niroj, next time try to do the bonus!

PR #3630, opened 2023-04-14
[Lis-wdft-2023] Lucas Telles

lltelles:


src/viking.js
```diff
@@ -1,11 +1,89 @@
...
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength){
+        super(health, strength)
+        this.name = name;
+    }
+    attack(){
+        return this.strength;
```

joaopdg (commented):
Great job on the lab and the bonus Lucas!

PR #3634, opened 2023-04-14
[LIS-WDFT-APR23] Pedro Nogueira

Pedro-No:


src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+        this.saxonArmy.push(Soldier);
+    };
+    vikingAttack() {
+        let randomVik = this.vikingArmy[Math.floor(Math.random()) * this.vikingArmy.length];
+        let randomSax = this.saxonArmy[Math.floor(Math.random()) * this.saxonArmy.length];
+
+        let attack = (randomSax.receiveDamage(randomVik.attack()));
+
+        if (randomSax.health <= 0) {
+            this.saxonArmy.pop(randomSax)
```
joaopdg, line 67: .pop() always removes the last element of the array, this is passing the test but you are not removing the Saxon that received the damage, for that you should have used the .splice()

src/viking.js
```diff
@@ -1,11 +1,94 @@
...
+        return attack;
+    };
+    saxonAttack() {
+        let randomVik = this.vikingArmy[Math.floor(Math.random()) * this.vikingArmy.length];
+        let randomSax = this.saxonArmy[Math.floor(Math.random()) * this.saxonArmy.length];
+
+        let attack = (randomVik.receiveDamage(randomSax.attack()));
+
+        if (randomVik.health <= 0) {
+            this.vikingArmy.pop(randomVik)
```
joaopdg, line 79: Same thing as before, use .splice()

joaopdg (commented):
Great job on the lab and the bonus Pedro!

PR #3635, opened 2023-04-14
lis-ft-apr23 Francisco Branco

chikifella7:


src/viking.js
```diff
@@ -1,11 +1,84 @@
...
+    this.saxonArmy.push(Saxon);
+  }
+  vikingAttack() {
+    let randomSaxon =
+      this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)];
+    let randomViking =
+      this.vikingArmy[Math.floor(Math.random() * this.saxonArmy.length)];
+    let deadSaxons = randomSaxon.receiveDamage(randomViking.attack());
+    if (randomSaxon.health <= 0) {
+      this.saxonArmy.splice(Math.floor(Math.random() * this.saxonArmy.length),1)
```
brunorocha20, line 62: You are removing a random saxon from the `saxonArmy` array, and writing the logic again, when you can remove the exact random saxon that you previously randomized its index. That `randomSaxon` variable, already has the index of your dead saxon, so, you should use that index to remove from the array.

brunorocha20 (commented):
Great job on the lab Francisco.

PR #3637, opened 2023-04-14
Lis-WDFT-APR-2023-Nuno-Durão

NunoDurao:


src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+        this.saxonArmy.push(saxon)
+    }
+    vikingAttack() {
+        let randomViking = this.vikingArmy[Math.floor(Math.random()) * this.vikingArmy.length];
+        let randomSaxon = this.saxonArmy[Math.floor(Math.random()) * this.saxonArmy.length];
+
+        let attack = (randomSaxon.receiveDamage(randomViking.attack()));
+
+        if (randomSaxon.health <= 0) {
+            this.saxonArmy.pop(randomSaxon)
```
joaopdg, line 70: .pop() always removes the last element of the array, this is passing the test but you are not removing the Saxon that received the damage, for that you should have used the .splice()

src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+        return attack;
+    };
+    saxonAttack() {
+        let randomViking = this.vikingArmy[Math.floor(Math.random()) * this.vikingArmy.length];
+        let randomSaxon = this.saxonArmy[Math.floor(Math.random()) * this.saxonArmy.length];
+
+        let attack = (randomViking.receiveDamage(randomSaxon.attack()));
+
+        if (randomViking.health <= 0) {
+            this.vikingArmy.pop(randomViking)
```
joaopdg, line 82: Same thing as before, use .splice() instead of .pop()

joaopdg (commented):
Great job on the lab and the bonus Nuno!

PR #3641, opened 2023-04-16
[LIS-WDFT-APR23] João Carrilho

jcarrilho:


src/viking.js
```diff
@@ -1,11 +1,107 @@
...
+    
+    battleCry() {
+    return "Odin Owns You All!";
+    }
+    }
+    
+    // Saxon
+    class Saxon extends Soldier {
+    constructor(health, strength) {
+    super(health, strength);
```
joaopdg, line 42: Since we are not passing a new parameter for the class Saxon we don't need the constructor neither the super:

```
class Saxon extends Soldier {
  receiveDamage(damage) {
    this.health -= damage;
    if (this.health > 0) {
    } else {
    }
  }
}
```

joaopdg (commented):
Great job on the lab and the bonus João, and good idea adding the extra methods in the War class!

PR #3645, opened 2023-04-20
[PTMEX042023]-Walter-Bernal

WaltBernalM:
This version of mine is in compliance with test plan and also bulletproof of some problems not covered related to possible bugs when simulating an actual battle with random properties in health, strength and an armies as big as 1k.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -120,28 +120,31 @@ Object.freeze(War)
 
 
 
-// ---------------------------------------------------------------- //
-// Debugging
-// Scope: uncovered areas of Jasmine automatical testing
+/* --------------------------- "Debugging" --------------------------- */
+// Scope: uncovered areas of Jasmine automatical testing (and have fun)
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,147 @@
...
+
+// Creation of random army
+for (let i = 0; i < 50; i++) {
+  const viking = new Viking(`viking #${i}`, getRandomNumber(), getRandomNumber())
+  const saxon = new Saxon(getRandomNumber(), getRandomNumber())
+  oldWar.addViking(viking)
+  oldWar.addSaxon(saxon)
+}
+
+// To the battle!
```
ta-web-mexpt2, line 139: nice comment here :metal: 

```suggestion
/**
* To the battle! 
*/
```

WaltBernalM: I've added a little bit of precautions just to avoid any undesired random war

PR #3646, opened 2023-04-20
[PTMEX042023]Victor-Tovar

TovCrowe:


src/viking.js
```diff
@@ -1,11 +1,100 @@
...
+        const randomSaxon = this.saxonArmy[randomSaxonArmy]
+
+        const resultBattle = randomViking.receiveDamage(randomSaxon.strength)
+
+        if(randomViking.health <= 0){
+            this.vikingArmy.splice(randomVikingArmy, 1)
+        }
+        return resultBattle
+    }
+
```
ta-web-mexpt2, line 96: `
    showStatus() {
        if (this.saxonArmy.length === 0) {
            return "Vikings have won the war of the century!";
        } else if (this.vikingArmy.length === 0) {
            return "Saxons have fought for their lives and survived another day...";
        } else {
        }
    }`

PR #3660, opened 2023-05-13
[FTRMT052023] Maria Friedemann

softcake1988:
lab done, bonus started but not finished - I got stuck a few times but managed to move on and learnt a lot while doing the lab. was cool

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,98 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+    }
+
+    // Iteration 1: Soldier
+attack() {
+    return this.strength;
+};
+
+receiveDamage(damage) {
+    this.damage = damage;
+    this.health = this.health - damage;
+    }
+};
 
 // Viking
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+    this.damage = damage;
+    this.health = this.health - damage;
+    }
+};
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+    super(name, health, strength);
```
TA-Remote, line 22: Here you would just pass what is needed for the parent's constructor 
```suggestion
    super(health, strength);
```

src/viking.js
```diff
@@ -1,11 +1,98 @@
...
+    this.strength = strength;
+    }
+
+    // Iteration 1: Soldier
+attack() {
+    return this.strength;
+};
+
+receiveDamage(damage) {
+    this.damage = damage;
```
TA-Remote, line 14: here you're setting the damage as this.damage, but later you're just using damage, so you could have skipped that line

TA-Remote (commented):
- Good work, keep practicing

PR #3664, opened 2023-05-13
viking lab done till bonus

olabonati:
this was fun but Jasmine can be frustrating
asside that not sure why i have to define  f. ex this.health or this.strength if theyr are already inherited from the parent??

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,81 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor (health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
 
+attack(){
+    return this.strength;
+}
+receiveDamage(damage){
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hi Ola, 
when we "extends" a Class we’re sure inheriting the parents properties and methods, to don’t declare again inside our constructor the same attribute we have to use the keyword ‘super’ and pass inside it the properties we are inheriting from the parent.   EX: 
```
//Parent Class
class Soldier {
    constructor(health, strength) {
        this.health = health;
        this.strength = strength;
    }
}

// Child Class
        super(health, strength); 
 // so that we can call this.health and this.strength (when we need them inside the Viking class) but we don’t need to declare them again
        this.name = name;
    }
}
```

🙂
Alessandra

PR #3666, opened 2023-05-13
(FT-Remote-WD-05.23) Paula Mateo

Anpal1986:
in the iteration 2: receiveDamage()
- should remove the received damage from the health property <-- I dont know how to do it to pass the test. I tried this: this.health-damage; but it seems to be wrong... could you help me out how to do it?

- Then when I need to return the messages in the class Sason --> receiveDamage (in the if clause): is returning the correct message but the test is as well failing... I dont know really why :s I tried, many different things, but nothing seems to work 
- 
thanks

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -25,10 +25,10 @@ class Viking extends Soldier {
       
         let status=this.health-damage;
         if(status>0){
-            return console.log(`"${this.name} has received ${damage} points of damage"`);
+            return `"${this.name} has received ${damage} points of damage"`;
         }
           else{
-            return console.log(`"${this.name} has died in act of combat`);
+            return `"${this.name} has died in act of combat`;
         }
     }
     battleCry(){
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,79 @@
...
+    
+    constructor(health, strength){
+        this.health=health;
+        this.strength=strength;
+        }
+    attack(){
+        return this.strength;
+            }
+    receiveDamage(damage){
+    this.health-damage;
```
TA-Remote, line 12: You forgot the `=`

```suggestion
    this.health -= damage;
```

src/viking.js
```diff
@@ -1,11 +1,79 @@
...
+    constructor(name, health, strength){
+        super(health,strength);
+        this.name=name;
+        
+                                       }
+    receiveDamage(damage){
+      
+        let status=this.health-damage;
+        if(status>0){
+            return `"${this.name} has received ${damage} points of damage"`;
```
TA-Remote, line 28: You have extra quotations here

```suggestion
            return `${this.name} has received ${damage} points of damage`;
```

src/viking.js
```diff
@@ -1,11 +1,79 @@
...
 
 // War
-class War {}
+class War {
+    constructor(){
+        this.vikingArmy=[];
+        this.saxonArmy=[];
+    }
+addViking(Viking){
+    Viking=this.vikingArmy[Viking];
```
TA-Remote, line 64:
```suggestion
addViking(viking) {
    this.vikingArmy.push(viking);
}
```

TA-Remote (commented):
- Good work
- Try to keep follow conventions to make easier for you and others to read your code (camelCase, indentation, names)

PR #3670, opened 2023-05-14
[FTRMT052023] Mary Sue Meissner

MarySueMe:


src/viking.js
```diff
@@ -1,11 +1,118 @@
...
+   
+}
+};
+
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength){
+        super('name');
```
TA-Remote, line 22: For the super we wanted to pass the information needed on the parent's constructor
```suggestion
        super(health, strength);
```

TA-Remote (commented):
- Good work, keep practicing

PR #3673, opened 2023-05-14
[FTRMT052023] Linh Nguyen

linhvnde:
Can you help to share the code snippet to test the lab with actual `new classes`? thank you

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,95 @@
 // Soldier
-class Soldier {}
+class Soldier {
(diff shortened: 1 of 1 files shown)


linhvnde:
for example I tried this and it didnt give me any result :(
`js
const mySoldier = new Soldier(1, 1);
const myViking = new Viking("Han", 1, 1);
const mySaxon = new Saxon();

const myTest = new War();`

PR #3674, opened 2023-05-14
[FTRMT052023] Pierre Docquin

Dokryptos:


src/viking.js
```diff
@@ -1,11 +1,124 @@
...
+        }
+    }
+}
+
 
 // War
-class War {}
+class War {
+    vikingArmy = [];
+    saxonArmy = [];
```
TA-Remote, line 64: Here I would place vikingArmy and and saxonArmy inside a `constructor()` and call it `this.vikingArmy` for example

src/viking.js
```diff
@@ -1,11 +1,124 @@
...
+        }
+        
+        return receiveDamageOfSaxon
+    }
+
+    saxonAttack(){
+        let randomSaxon = this.saxonArmy[Math.floor(Math.random() * (this.saxonArmy.length))] ;
+        let randomVikingInd = Math.floor(Math.random() * (this.vikingArmy.length));
+        let randomViking = this.vikingArmy[randomVikingInd];
+        let receiveDamageOfViking = this.randomViking.receiveDamage(this.randomSaxon.strength);
```
TA-Remote, line 89: Here you need to remove the `this` from `randomViking` and `randomSaxon` because you're creating those inside the method, so you don't need to access them with the `this` keyword, in that case the method is trying to access those variable from the object (war) scope and not the ones that you're creating inside the method

TA-Remote (commented):
- Good work

PR #3682, opened 2023-05-14
 [FTRMT052023]  Lucas Gavião

Lucas-Gaviao:


src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+
 
 // War
-class War {}
+class War{
+    constructor(){
+
+    }
+    addViking(Viking){
+        vikingArmy = [];
```
TA-Remote, line 58: I would put the armies in the constructor using the `this` keyword

src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+        vikingArmy.push(Viking);
+    }
+    addSaxon(Saxon){
+        saxonArmy = [];
+        saxonArmy.push(Saxon);
+    }
+    vikingAttack(){
```
TA-Remote, line 65: - In the attacks you need to select one attacker randomly
- Select one defender (also randomly)
- Check if the defender is alive
- If the defender died removed it from the army's arrays

src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+        saxonArmy.receiveDamage();
+    }
+
+    saxonAttack(){
+        vikingArmy.receiveDamage();
+    }
+    showStatus(){
```
TA-Remote, line 72: This one possibly is correct, but you need to implement the attacks to be able to pass the tests

TA-Remote (commented):
- Good work, keep practicing :)

PR #3729, opened 2023-07-14
[LIS-FT-JULY23] Gonçalo Matias

Gosamat:


jacobviv:
Very good job Gonçalo!

src/viking.js
```diff
@@ -1,11 +1,103 @@
...
+        if (this.vikingArmy.length === 0){
+
+            return "Saxons have fought for their lives and survived another day...";
+        }
+        else if(this.saxonArmy.length === 0){
+            return "Vikings have won the war of the century!";
+
+        }
+        else if(this.vikingArmy.length > 0 && this.saxonArmy.length > 0){
+            return "Vikings and Saxons are still in the thick of battle.";
```
jacobviv, line 97: Gonçalo, here you can use the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for the remaining answer.

PR #3732, opened 2023-07-14
[LISFTJULY23] Julia Guillotte

jsguillotte:


src/viking.js
```diff
@@ -1,11 +1,107 @@
...
+    
+    showStatus() {
+        if(!this.saxonArmy.length && this.vikingArmy.length) {
+            return `Vikings have won the war of the century!`
+        }
+        
+        else if(!this.vikingArmy.length && this.saxonArmy.length) {
+            return `Saxons have fought for their lives and survived another day...`
+        }
+        else if(this.vikingArmy.length>0 && this.saxonArmy.length > 0) {
```
brunorocha20, line 97: Julia, here, you can get used to the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for the remaining answer.

brunorocha20 (commented):
Great job on the lab Julia.

PR #3733, opened 2023-07-14
[LISFTJULY23] Rita Saraiva

Rtzz123:


src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+    }
+    return result;
+  }
+
+  showStatus() {
+    if (this.saxonArmy.length === 0) {
+      return 'Vikings have won the war of the century!';
+    } else if (this.vikingArmy.length === 0) {
+      return 'Saxons have fought for their lives and survived another day...';
+    } else if (this.saxonArmy.length >= 1 && this.vikingArmy.length >= 1) {
```
brunorocha20, line 98: Here, you can get used to the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for your last and remainder option.

brunorocha20 (commented):
Great job on the lab.

PR #3734, opened 2023-07-14
[LISFTJULY23] FILIPA FLORA LAB5

filipaflora:


src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+    
+    showStatus(){
+        if (this.saxonArmy.length === 0 ){
+            return 'Vikings have won the war of the century!'
+        } 
+        else if (this.vikingArmy.length === 0){
+            return "Saxons have fought for their lives and survived another day..."
+        }
+        else if(this.saxonArmy.length >= 1 && this.vikingArmy.length >= 1){
+            return "Vikings and Saxons are still in the thick of battle."
```
jacobviv, line 101: Filipa, here you can use the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for the remaining answer.

jacobviv:
Very good lab Filipa!

src/viking.js
```diff
@@ -1,11 +1,105 @@
...
+        }
```
jacobviv, line 102: Filipa, here you can use the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for the remaining answer.

PR #3739, opened 2023-07-14
[LIS-FT-JULY23] Tomás Borda de Água

Vayre047:


src/viking.js
```diff
@@ -1,11 +1,88 @@
...
+        this.strength = strength;
+        this.damage = 0;
+    }
+
+    attack(){
+        return this.strength;
+    };
+
+    receiveDamage(damage){
+        this.damage = damage;
```
brunorocha20, line 14: Here, `damage` is going to be the argument of your `receiveDamage()` method, so, you don't need to reference it to the `class` as you are receiving some value (later on) and that's the value you are going to use to remove it from the `health`.

src/viking.js
```diff
@@ -1,11 +1,88 @@
...
+    saxonAttack(){
+
+    }
+
+    showStatus(){
+        if(this.saxonArmy.length === 0){
+            return "Vikings have won the war of the century!"
+        }else if(this.vikingArmy.length === 0){
+            return "Saxons have fought for their lives and survived another day...";
+        }else if(this.saxonArmy.length && this.vikingArmy.length ){
```
brunorocha20, line 84: Here, you can get used to the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for your last and remainder option.

brunorocha20 (commented):
Good job on the lab.

PR #3742, opened 2023-07-14
 [LIS-FT-JULY23] João_Borrega

JoaoBorrega:


jacobviv:
Good job with your lab João

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health, strength);
+        this.name = name;
+        this.health = health;
+        this.strength = strength;
```
jacobviv, line 22: Since you extended Viking from Soldier, and on your `super` you inherit `health` and `strength`, you don't need to pass their values on the constructor. It could be written like this:

```
    super(health, strength);
    this.name = name;
  }
(...)
}
```

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+    }
+    attack() {
+        return this.strength
+    }
+    battleCry() {
+        let odim = "Odin Owns You All!";
+        return odim;
+    }
+    receiveDamage(damage) {
+        this.damage = damage;
```
jacobviv, line 32: Here, `damage` is going to be the argument of your `receiveDamage()` method, so, you don't need to reference it to the class as you are receiving some value (later on) and that's the value you are going to use to remove it from the `health`.

PR #3749, opened 2023-07-14
[LIS-FT-JULY23] mariana_silva

maryanasilva:


src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+    receiveDamage(damage){
+        this.health -= damage;
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength){
+        super(name, health, strength);
```
brunorocha20, line 18: Since you extended Viking from Soldier, and on your `super` you inherit `health` and `strength`, you don't need to pass their values on the constructor. It could be written like this:
```
    super(health, strength);
    this.name = name;
  }
(...)
}
```

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+        this.name = name;
+        this.health = health;
+        this.strength = strength;
+    }
+    attack(){
+        return this.strength;
+    }
+    receiveDamage(damage){
+        this.damage = damage;
```
brunorocha20, line 27: Here, `damage` is going to be the argument of your `receiveDamage()` method, so, you don't need to reference it to the class as you are receiving some value (later on) and that's the value you are going to use to remove it from the `health`.

src/viking.js
```diff
@@ -1,11 +1,102 @@
...
+    }
+
+    showStatus(){
+        if(this.saxonArmy.length === 0){
+            return 'Vikings have won the war of the century!';
+        }
+        else if(this.vikingArmy.length === 0){
+            return 'Saxons have fought for their lives and survived another day...';
+        }
+        else if(this.saxonArmy.length === 1 && this.vikingArmy.length === 1){
```
brunorocha20, line 98: Here, you can get used to the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for your last and remainder option.

brunorocha20 (commented):
Good job on the lab Mariana.

PR #3750, opened 2023-07-14
[LIS-FT-072023] João Elias

johnseliseto:


src/viking.js
```diff
@@ -1,11 +1,106 @@
...
+      this.vikingArmy.splice(randomViking, 1);
+    }
+    return fight2;
+  }
+  showStatus() {
+    if (this.saxonArmy.length === 0) {
+      return "Vikings have won the war of the century!";
+    } else if (this.vikingArmy.length === 0) {
+      return "Saxons have fought for their lives and survived another day...";
+    } else if (this.saxonArmy.length === 1 && this.vikingArmy.length === 1) {
```
jacobviv, line 102: João, here you can use the `else`, since you targeted the two other options with the `if` and the `else if`, you can trigger the `else` operator for the remaining answer.

PR #3751, opened 2023-07-16
[LISFTJULY23] Mariana Marques

Haesonni:


src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+  addSaxon(Saxon) {
+    this.saxonArmy.push(Saxon);
+  }
+  vikingAttack() {
+    let randomSaxon = this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)];
+    let randomViking = this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
+
+    let result = randomSaxon.receiveDamage(randomViking.strength);
+
+    for (let i = 0; i < this.saxonArmy.length; i++) {
```
brunorocha20, line 65: Here, you are running a loop to check if there's "dead" saxons and removing them from the array. The thing is, you are applying a `vikingAttack()` method, to target a random saxon, but since you have it's random index, it will be that same index you'll use to check if the `health` is smaller or equal to `0` and for you to remove it from the `saxonArmy` array. 
So, there's no need for this 'extra' step of going trough all the array and finding the ones who are dead.

brunorocha20 (commented):
Good job on the lab Mariana.

PR #3752, opened 2023-07-16
[LIS-FT-JULY23] maureen_treacy

maureen-treacy:


src/viking.js
```diff
@@ -1,11 +1,82 @@
...
+    receiveDamage(damage) {
+        this.health = this.health - damage
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier { 
+    constructor(name, health, strength) {
+        super(health, strength) //Question for T.A.: does super always come before the arguments? When i tried switching this super line and the one after, it wasn't correct on Jasmine
```
brunorocha20, line 18: Yes, that's the order you should always use. 
`class`, `constructor`, and, if needed `super`

brunorocha20 (commented):
Good job on the lab Maureen

PR #3754, opened 2023-07-17
[LIS-FT-JULY23] Miguel Leite

AakaiLeite:


src/viking.js
```diff
@@ -1,11 +1,91 @@
...
+  }
+  vikingAttack() {
+    let randomViking =
+      this.vikingArmy[Math.floor(Math.random() * this.vikingArmy.length)];
+    let randomSaxon =
+      this.saxonArmy[Math.floor(Math.random() * this.saxonArmy.length)];
+    let damage = randomViking.strength;
+    let fight = randomSaxon.receiveDamage(damage);
+    if (randomSaxon.health <= 0) {
+      this.saxonArmy.pop(randomSaxon);
```
webLisbonTA, line 66: Miguel, here, using the `.pop()` method to remove your dead saxon from the array, always removes the last index of the array even if you pass a determined index to it. In this case, Jasmine is accepting that answer, but the most correct way to do that would be using the `.splice()` method, where you can define an index to start removing and the amount that you want to remove from the array.
So, it would be something like this:
```
    }
```

webLisbonTA (commented):
Great job on the lab.

PR #3762, opened 2023-08-04
[FTRMT072023] Priscila

Priscila-Borges:
I didn't manage to do the "War" iteration :(

[Click here to see my current status!](https://tenor.com/view/confuses-handsup-dontknow-hands-up-gif-24764458)

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,85 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
+        this.strength = strength;
+    }
+
+    attack() {
+        return (this.strength);
+    }
+
+    receiveDamage(theDamage) {
+        this.health = this.health - theDamage;
+    }
+}
+
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength) {
+        super(health, strength);
+        this.name = name;
+    }
+
+    receiveDamage(theDamage) {
+        this.health = this.health - theDamage;
+        this.theDamage = theDamage
+       
+        if (this.health > 0) {
+            return `${this.name} has received ${this.theDamage} points of damage`;
+        }
+        else {
+            return `${this.name} has died in act of combat`;
+        }
+    }
+
+    battleCry(){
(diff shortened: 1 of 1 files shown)


TA-Remote:
hey @Priscila-Borges  
well done with your solution🙌
About the saxon and viking attack methods, hope that after this morning lab review are more clear, but in any case I'll share some notes here😉 (same logic will apply for the saxonAttack)

`
```
  vikingAttack() {
    // generate a random index number between 0 and the length of the vikingArmy array
    // and store it in a variable
    let randomVikingIndex = Math.floor(Math.random() * this.vikingArmy.length); //  give a number
    // use the index to pick a viking from the vikingArmy
    let randomViking = this.vikingArmy[randomVikingIndex]; //  a viking object
    // do the same but with the saxonArmy
    let randomSaxon = this.saxonArmy[randomSaxonIndex]; //  a saxon object
    // Store in a variable the result of calling the method "receiveDamage" on the saxon, 
    // passing the strength of the viking as argument or --> randomViking.attack() 
    // --> the attack() method of the soldier, return the strength
```
    
```
    // the receiveDamage method:
    // 1. It's waiting to receive the damage as a parameter (the strength)
    // 2. it's subtracting the damage (the strength) 
    // from the health of the soldier we called the method on (randomSaxon)
    // 3. It's checking if the soldier is still alive or dead, returning a message
    let thingToReturnAtTheEnd = randomSaxon.receiveDamage(randomViking.attack());
```

```
    // Now that the property health of the saxon has been updated after receiving the damage from the viking,
    // we want to remove the saxon object from the saxonArmy array, if it's dead, so if the health is less or equal to 0

      // to remove it we use the .splice() method
      // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice
      this.saxonArmy.splice(randomSaxonIndex, 1);
    }
```
    
```
    // we return the result we previously stored: it'll print the message of the 
    // randomSaxon.receiveDamage(randomViking.attack()) in the console
    return thingToReturnAtTheEnd;
  }
```
`
If you have any question, please let us know🙂

keep it up💪

Ale

PR #3763, opened 2023-08-05
[FTRMT072023]Anna

AnnaKazulak:
Hi Ale. I solved also the bonus part. For the vikingAttack methode I used help form ChatGPT. I knew I need to use Math.random() method but I was getting confused and it was late at night. To improve the code I should have written helpe functions like e.g. getRandomIndex() to find a random solder and removeDeatSoldier() to remove the ones who  "has died in combat"
 and call this functions in VikingsAttack() and SaxonAttack().
 I was getting confused bei improving the code, so I left it in no DRY option. At least code is working properly ;)

Have a  nice weekend!!

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,116 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+  receiveDamage(damage) {
+    this.damage = damage;
+    this.health -= this.damage;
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hi @AnnaKazulak 
Well done with your no DRY solution😉🙌
It's totally fine to use the tools we have to get hints, expecially when we don't remember the correct syntax!
About your idea to refactor the code... whenever you spot repetitions, go step by step ( starting with getRandomSoldiers would be a nice one😜) you'll have a loooot of opportunity to practice!
 Keep it up💪

PR #3772, opened 2023-08-06
[FTRT072023]Hanna

HannaInIT:
I did only a part of homework


TA-Remote:
hey @HannaInIT 
you really did a good job with your solution and the code you produced🙌
Hope the concept of class is clear, and if you still have any doubt just let us know😉
Ale

PR #3855, opened 2023-10-20
Done with few errors

EminmertToprak:
In the bonus part, for vikingAttack, saxonAttack and showStatus, I asked help of AI and even though I didn't get the full answer from it and I understood the nature of the code, I'm not super confident about it. Also I had errors for recieving damages and returning strings for both sides even though my codes seems about right.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,103 @@
 // Soldier
-class Soldier {}
+class Soldier {
+	constructor(health, strength) {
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Mert. Don't worry, there are a few ways to accomplish these functions even though the tests are not turning green, we are working on that 😉 Great job on this lab!

PR #3872, opened 2023-10-22
i had a hard time finishing the exercise even thouhg pool gave us a h…

camilaflorez-dev:
…int on how to finish 'war' getting a random element in an array, i didn't fully understand,jasmine just keep giving me error so i skipped the bonus, Luis comment about using concatenation operators helped me understand the code better, and the code formatting tool also helped. I still feel very unsure with the exorcises

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,69 @@
-// Soldier
-class Soldier {}
+/* Soldier
+class Soldier {}*/
 
-// Viking
-class Viking {}
+class Soldier {
+    constructor(health, strength) {
+        this.health = health;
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey @camilaflorez-dev 
well done with your solution🙌

Hope this morning lab review  was helpful to have an idea on how to procede working on the War class💪
Jasmine is a real complainer! but don't worry to leave  your code and what you were trying (maybe commented), in this way we can take a look and go through your logic together😉
Sure Prettier is a life safer!

Keep it up
Ale🙂

PR #3890, opened 2023-11-30
Iterations 0 to 3 done

Natpinper:
Working on Bonus, trying to figure out vikingAttack() method

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,72 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health=health
+        this.strength=strength
+    }
+    attack(){
+        return this.strength
+    }
+    receiveDamage(damage){
+        this.health-= damage
+
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name,health, strength){
+        super(health, strength)
+        this.name=name
+    }
+    receiveDamage(damage){
+        this.health-= damage
+        if(this.health>0){
+            return `${this.name} has received ${damage} points of damage`
+        }else {
+            return `${this.name} has died in act of combat`
+        }
+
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,72 @@
...
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name,health, strength){
+        super(health, strength)
+        this.name=name
+    }
+    receiveDamage(damage){
+        this.health-= damage
+        if(this.health>0){
```
Web-Dev-PT-EN, line 24: I would use a little bit more spacing here, just to make the code easier to read
```suggestion
        if (this.health>0) {
```

src/viking.js
```diff
@@ -1,11 +1,72 @@
...
+            return `${this.name} has received ${damage} points of damage`
+        }else {
```
Web-Dev-PT-EN, line 26: The same as the previous, but also to keep the pattern, as you added a space after the `else`
```suggestion
        } else {
```

src/viking.js
```diff
@@ -1,11 +1,72 @@
...
+            return `${this.name} has died in act of combat`
+        }
+
```
Web-Dev-PT-EN, line 29: I would remove that blank line at the end of the method
```suggestion
```

src/viking.js
```diff
@@ -1,11 +1,72 @@
...
+    }
```
Web-Dev-PT-EN, line 30: I would also add a line after the method so you have more visual space, so a little easier to understand the different methods
```suggestion
    }
    
```

Web-Dev-PT-EN (commented):
- Good work, if possible I would suggest you to also work on the War iteration as it could be really helpful for the project

- Try to keep your indentantion and code organization consistent whenever is possible :)

PR #3892, opened 2023-12-01
[WD-PT] Nadine Rangosch

nenuphars:
I had some issues when playing the game and not making showStatus along the way. It would give me an error about not being able to read the health during the receiveDamage function and I think it might be that there were no soldiers left in the army at that time, so therefore there were no health points to be read? I was a little confused about the conditions in the showStatus and the way I did it now, might not be the best.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,97 @@
 // Soldier
-class Soldier {}
+class Soldier {
+  constructor(health, strength) {
+    this.health = health;
+    this.strength = strength;
+  }
+  attack() {
+    return this.strength;
+  }
+  receiveDamage(damage) {
+    this.health -= damage;
+  }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier {
+  constructor(name, health, strength) {
+    super(health, strength);
+    this.name = name;
+  }
+  receiveDamage(damage){
+    super.receiveDamage(damage)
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,182 @@
...
+        return "Vikings and Saxons are still in the thick of battle.";
+      }
+    else if (this.saxonArmy.length === 0) {
+      return "Vikings have won the war of the century!";
+    }
+    else if (this.vikingArmy.length === 0) {
+      return "Saxons have fought for their lives and survived another day...";
+    }
+  }
+  constructor() {
```
Web-Dev-PT-EN, line 109: I think it doesn't matter for the constructor to be at the end, but seems a little strange hehe

Web-Dev-PT-EN (commented):
Good work

- To make the showStatus show in the terminal you need a `console.log(battle.showStatus())` remember that the method is just returning the message, so maybe at some point one of the armies are empty and you're trying to execute an attack
image

The console.log appears that way because of the console ninja extension

src/viking.js
nenuphars: Haha yes, I think I did it because I was paying too much attention to the order in which it was explained in the instructions, but I can see how it makes sense to put it at the beginning

PR #3893, opened 2023-12-01
WD PT Eleanor Pitcher

eleanorpitcher:


src/viking.js
```diff
@@ -1,11 +1,65 @@
...
-class Viking {}
+class Viking extends Soldier {
+    constructor(name, health, strength){
+        super(health, strength)
+
+        this.name = name;
+    }
+    receiveDamage(damage){
+        this.health -= damage;
+        if (this.health > damage){
```
Web-Dev-PT-EN, line 24: Here we should check if the viking survived after receiving the damage
```suggestion
        if (this.health > 0){
```

src/viking.js
```diff
@@ -1,11 +1,65 @@
...
+        return "Odin Owns You All!"
+    }
+}
 
 // Saxon
-class Saxon {}
+class Saxon extends Soldier {
+    receiveDamage(damage){
+        this.health -= damage;
+        if (this.health > damage){
```
Web-Dev-PT-EN, line 39: The same as the previous :) On the vikings one we have a false positive for the test actually, as it seems
```suggestion
        if (this.health > 0){
```

Web-Dev-PT-EN (commented):
If you have the time I would suggest you to implement the war iteration, it could be helpful for the project too :)

PR #3894, opened 2023-12-01
[WD-PT] Eduardo Botelho

UPSxACE:


Web-Dev-PT-EN (commented):
Very good lab, congratulations :) The random function could also be a method on the War, so it's not on the global context

PR #3929, opened 2024-01-26
[BERLIN-PT-012024]Franciely Beckert

FranBeckert:
Jasmine keeps saying there is something wrong with vikingsAttack() method, but I don't see it.

### DIFF
--- a/src/viking.js
+++ b/src/viking.js
@@ -1,11 +1,99 @@
 // Soldier
-class Soldier {}
+class Soldier {
+    constructor(health, strength) {
+        this.health= health;
+        this.strength = strength;        
+    }
+    attack() {
+        return this.strength;
+    }
+
+    receiveDamage(damage) {
+        this.health -= damage;
+    }
+}
 
 // Viking
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength) {
+        super(health, strength)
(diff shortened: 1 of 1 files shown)


src/viking.js
```diff
@@ -1,11 +1,99 @@
...
-class Viking {}
+class Viking extends Soldier{
+    constructor(name, health, strength) {
+        super(health, strength)
+        this.name = name;
+    }
+    receiveDamage(damage) {
+        this.health -= damage;        
+        if (this.health) {
+                return  `${this.name} has received ${damage} points of damage`
```
jacked97, line 25: wrong intendation here

src/viking.js
```diff
@@ -1,11 +1,99 @@
...
+        let result = randomViking.receiveDamage(randomSaxon.strength);
+
+        if (randomViking.health <= 0) {
+            this.vikingArmy.splice(randomVikingIndex, 1);
+        }
+        return result;
+    }
+    showStatus () {
+        if (this.saxonArmy.length <= 0) {
+            return `Vikings have won the war of the century!`;
```
jacked97, line 92: move the constants (values that stays the same to the top of the file)

jacked97 (changes requested):

